PR #15, opened 2024-05-25
Do not run nightly workflows on forks.

advaitjain:
This change can't really be tested until:

 * it is merged: to make sure nothing breaks for the upstream repo
 * forks update to upstream/main: to confirm that workflows are not triggered in forks.

BUG=cleanup github workflows.

### DIFF
--- a/.github/workflows/mark_stale.yml
+++ b/.github/workflows/mark_stale.yml
@@ -12,6 +12,9 @@ on:
 
 jobs:
   stale:
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'schedule' && github.repository == 'google-ai-edge/ai-edge-torch') # don't run in forks.
 
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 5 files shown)


chunnienc (approved):

chunnienc:
I thought workflows are disabled by default in github forked repo, and fork owner can disable workflows in their settings. Maybe better for me to understand what issues you saw for this.

advaitjain:
I should have provided more context with the PR.

Workflows appear to be enabled by default in forks. For example: https://github.com/advaitjain/ai-edge-torch/actions

Anyone forking does have the option of disabling then, but that isn't great either. It's an additional step that can easily be overlooked (as I did). And I might want to test some workflow in my fork.

The approach from this PR is what we use with TensorFlow as well.

chunnienc:
The message I saw in the actions page in my fork repo is this though (https://github.com/chunnienc/ai-edge-torch/actions):

> Workflows aren’t being run on this forked repository
> Because this repository contained workflow files when it was forked, we have disabled them from running on this fork. Make sure you understand the configured workflows and their expected usage before enabling Actions on this repository.

I think this change is fine as it won't affect our dev flow and tensorflow uses this approach. Let's proceed with this.

PR #17, opened 2024-05-28
Init nightly release CI (1/2) 

chunnienc:
Triggered after unit-test nightly succeeds. The initial version publishes the nightly to pypi ai-edge-torch-nightly.

Release preview (in my fork): https://github.com/chunnienc/ai-edge-torch/releases/tag/v0.2.0.dev20240528

Upcoming PRs:
- Refactor nightly tests to run before/with nightly releases

BUG=https://b.corp.google.com/issues/342188725


.github/workflows/nightly_release.yml
```diff
@@ -0,0 +1,69 @@
...
+          echo "version=${VERSION}" >> $GITHUB_OUTPUT
+
+      - name: Upload release to github
+        uses: softprops/action-gh-release@v2
+        with:
+          files: |
+            dist/*.whl
+            requirements.txt
+          prerelease: true
+          tag_name: v${{ steps.wheel_version.outputs.version }}
```
advaitjain, line 69: I was wondering if we should skip uploading the nightly artifacts on github, and instead only have the nightly pip package. And keep https://github.com/google-ai-edge/ai-edge-torch/releases for manual releases only.

Likely worth a chat before we decide one way or the other.

chunnienc: Releasing to pypi only should be fine. I set this up while considering to keep requirements.txt in github releases, but that should be minor and we won't expect people to install old nightly versions.

chunnienc: Updated to upload the nightly to pypi. Please take a look again.

advaitjain (approved):

PR #24, opened 2024-05-31
Create attention layers with external kv cache and prepare an example.

haozha111:
This PR contains following changes:
1) Fork the current attention code into layers/experimental/attention.py, which supports external KV cache style (for GPU work). This will avoid polluting the current internal-kv attention implementation.
2) Put SDPA functions into separate files to allow code reuse.
3) Prepare a toy model w/ 2 transformer blocks, using external kv cache layers.

Note conversion to TFL fails because tfl's dynamic update slice doesn't support int64 index, will work on a fix later.

BUG=b/338453350


ai_edge_torch/generative/layers/experimental/attention.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+# CausalSelfAttention which can support MHQ, MQA or GQA.
+class CausalSelfAttention(nn.Module):
+
+  def __init__(
+      self,
+      dim: int,
+      config: cfg.AttentionConfig,
+      kv_cache_max: int,
```
hheydary, line 103: This argument does not seem to be used anywhere.

ai_edge_torch/generative/examples/test_models/toy_model_with_external_kv_cache.py
```diff
@@ -0,0 +1,162 @@
...
+  ) -> (torch.Tensor, torch.Tensor, torch.Tensor):
+    x = self.tok_embedding(idx)
+    cos, sin = self.rope_cache
+    cos = cos.index_select(0, input_pos)
+    sin = sin.index_select(0, input_pos)
+    mask = self.mask_cache.index_select(2, input_pos)
+    mask = mask[:, :, :, : self.config.max_seq_len]
+
+    for i, block in enumerate(self.transformer_blocks):
+      input_k = torch.clone(k_caches[i])
```
hheydary, line 76: Why are we cloning these tensors?

haozha111: good catch, we don't need those. originally i thought this was needed to walk-around a dynamo export issue, but trying again seems it works even without this hack.

ai_edge_torch/generative/layers/experimental/attention.py
haozha111: removed.

ai_edge_torch/generative/layers/experimental/attention.py
```diff
@@ -0,0 +1,187 @@
...
+      x_norm = self.pre_atten_norm(x)
+      attn_out, kv_cache = self.atten_func(x_norm, rope, mask, input_pos, kv_cache)
+      x = x + attn_out
+      x_norm = self.pre_ff_norm(x)
+      output = x + self.ff(x_norm)
+
+    return output, kv_cache
+
+
+# CausalSelfAttention which can support MHQ, MQA or GQA.
```
hheydary, line 95: This could be added as class docstring instead of inline comment.

hheydary (dismissed):

ai_edge_torch/generative/layers/experimental/attention.py
haozha111: cool, original layer updated as well. now we need to keep both in sync to ensure a smooth migration.

hheydary (approved):

PR #27, opened 2024-06-03
Unpin tf-nightly in requirements.txt

chunnienc:
ai-edge-torch-nightly will always pull in the latest tf-nightly. This will allow bug fixes and improvements to the TFLite Converter to be automatically available for ai-edge-torch without the need to manually bump up the tf-nightly version. (from @advaitjain)

This PR also disables pip cache in CIs (python unit tests and model coverage) to always collect latest tf-nightly. If regression happens due to tf-nightly, the failures would be captured in nightly readme badges and PR pre merge checks.

BUG=none

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ numpy
 tabulate
 safetensors
 --pre
-tf-nightly==2.17.0.dev20240529
+tf-nightly>=2.17.0.dev20240529
 -f https://download.pytorch.org/whl/nightly/torch_nightly.html
 torch==2.4.0.dev20240429+cpu
 -f https://download.pytorch.org/whl/nightly/torch_nightly.html
--- a/.github/workflows/model_coverage.yml
+++ b/.github/workflows/model_coverage.yml
@@ -38,7 +38,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - run: python -m pip install --upgrade pip setuptools
+      - run: |
+          python -m pip install --upgrade pip setuptools
+          python -m pip cache purge
 
       - name: Setup benchmark repository
         run: |
--- a/.github/workflows/model_coverage.yml
+++ b/.github/workflows/model_coverage.yml
@@ -37,8 +37,6 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-          cache: "pip"
-          cache-dependency-path: "**/*requirements.txt"
 
       - run: python -m pip install --upgrade pip setuptools
 
--- a/.github/workflows/unittests_python.yml
+++ b/.github/workflows/unittests_python.yml
@@ -31,7 +31,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - run: python -m pip install --upgrade pip setuptools
+      - run: |
+          python -m pip install --upgrade pip setuptools
+          python -m pip cache purge
 
       - name: Install dependencies
         run: |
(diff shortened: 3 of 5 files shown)


requirements.txt
```diff
@@ -3,7 +3,7 @@ numpy
 tabulate
 safetensors
 --pre
-tf-nightly==2.17.0.dev20240529
+tf-nightly>=2.17.0.dev20240529
```
advaitjain, line 6: is the version number needed or can it be `tf-nightly` or perhaps `tf-nightly?=2.17`?

advaitjain (commented):
Could you add to the PR description what this change will do?

My assumption is the following (please correct if I got this wrong):

 * ai-edge-torch-nightly will always pull in the latest tf-nightly. This will allow bug fixes and improvements to the TFLite Converter to be automatically available for `ai-edge-torch` without the need to manually bump up the `tf-nightly` version.

Some downsides:
 * if there is a change in tf-nightly that breaks ai-edge torch then it will also break anyone who is developing from the ai-edge-torch repo locally and happens to have run a `pip install -r dev-requitrements.txt` when `tf-nightly` broke `ai-edge-torch`
 * Note that the build badge will also reflect this state -- which build badge will show this? unit tests?

We expect `tf-nightly` breaking `ai-edge-torch` to be infrequent and the benefits of automatically getting TFLite Converter improvements to outweigh the downsides.

requirements.txt
chunnienc: We make changes in ai-edge-torch that depends on corresponding changes in tflite converter e.g. lowering of  composite, so `>=` is needed.

chunnienc:
> * Note that the build badge will also reflect this state -- which build badge will show this? unit tests?

This part is a bit complicated. We have the pip cache enabled for CIs so it won't always use the latest tf-nightly for testing. Only when the soft pin (>=ver) is updated the cache would be invalidated.

This PR somehow assumes that ai-edge-torch almost never fails with latest tf-nightly. If a regression happens with latest tf-nightly, we will then need to update the soft pin back to a `==` hard pin until the fixes are landed in tf-nightly. For such case I'd rather have `==` in requirements.txt and encourage updating it more frequently (or setup CIs to do so).

advaitjain:
> > * Note that the build badge will also reflect this state -- which build badge will show this? unit tests?
> 
> This part is a bit complicated. We have the pip cache enabled for CIs so it won't always use the latest tf-nightly for testing. Only when the soft pin (>=ver) is updated the cache would guarantee to be invalidated.
> 
> This PR somehow assumes that ai-edge-torch almost never fails with latest tf-nightly. If a regression happens with latest tf-nightly, we will then need to update the soft pin back to a `==` hard pin until the fixes are landed in tf-nightly. For such case I'd rather have `==` in requirements.txt and encourage updating it more frequently (or setup CIs to do so with model coverage and unit tests).

Understood and I'm also comfortable with this assumption for now.

What I would still like to have is a build badge that shows when this breakage happens so that we can avoid a situation where folks see errors locally without a way to verify if our CI is also showing a regression. Is there something we can do on this front?

chunnienc:
> What I would still like to have is a build badge that shows when this breakage happens so that we can avoid a situation where folks see errors locally without a way to verify if our CI is also showing a regression. Is there something we can do on this front?

Minimal: disable pip cache for all nightly CIs to use latest tf-nightly, where the status are shown in readme badges. 
Full: disable pip cache in all CIs. PRs may fail surprisingly if tf-nightly regression happens. (Notice that working with latest tf-nightly does not require this since updating the soft pin also invalidates the cache).

advaitjain (approved):

PR #29, opened 2024-06-03
Wrap GELU in composite so that it is lowered to tfl.gelu

majiddadashi:
BUG=343777419


chunnienc (approved):
LGTM, thanks. Please merge after the lowering is ready in tf-nightly.

chunnienc (approved):

PR #30, opened 2024-06-04
Refactor layers for CLIP text encoder of SD model

yichunk:
Refactor layers for CLIP text encoder of SD model. Tested to successfully generate a proper image as before refactoring.

There are several updates in this PR.

1. Clean the split UNet implementation in SD
2. Refactor CLIP to use the layers module
3. ~~Add `GATED_SHARED` type in `FeedForwardType`, which is used in CLIP.~~
4. Add `qkv_transpose_before_split` filed in `AttentionConfig`, which is used in CLIP.
5. Add `GELU_QUICK` type in `ActivationType`, which is used in CLIP.
6. Add `attn_fused_qkv_proj` in `TensorNames`, which bundle qkv projection tensors in one tensor, used in CLIP.
7.  Add `embedding_position` in `TensorNames`, which is a learned position embedding, used in CLIP.

BUG=b/311216181


ai_edge_torch/generative/utilities/loader.py
```diff
@@ -158,7 +170,7 @@ def load(self, model: torch.nn.Module, strict: bool = True):
       raise ValueError(
           f"Failed to map all tensor. Remaing tensor are: {list(state.keys())}"
       )
-    model.load_state_dict(converted_state, strict=strict)
+    return model.load_state_dict(converted_state, strict=strict)
```
hheydary, line 173: Please update docstring and pytype annotations.

ai_edge_torch/generative/layers/attention.py
```diff
@@ -159,28 +159,35 @@ def forward(
...
+      qkv = qkv.view(
+          B, T, total_qkv, self.config.num_query_groups, self.head_dim
+      )  # (B, T, total_qkv, num_query_groups, head_dim)
+      q, k, v = qkv.split((q_per_kv, 1, 1), dim=-3)
+    else:
+      qkv = qkv.view(
+          B, T, self.config.num_query_groups, total_qkv, self.head_dim
+      )  # (B, T, num_query_groups, total_qkv, head_dim)
+      # Split batched computation into three.
+      q, k, v = qkv.split((q_per_kv, 1, 1), dim=-2)
```
haozha111, line 173: it seems the only difference between the two cases is the order of `total_qkv` in the first view operation. After split, the q/k/v tensors should be same for both cases? If so, can we just keep one case here?

ai_edge_torch/generative/layers/builder.py
```diff
@@ -97,6 +99,8 @@ def _get_activation(type_: cfg.ActivationType):
     return F.gelu
   elif type_ == cfg.ActivationType.GELU_TANH:
     return lambda x: F.gelu(x, approximate="tanh")
+  elif type_ == cfg.ActivationType.GELU_QUICK:
```
haozha111, line 102: add some comment about GELU_QUICK here?
e.g. https://github.com/huggingface/transformers/blob/8685b3c5d2dd2550527773d2a02499495a759e31/src/transformers/activations.py#L92

ai_edge_torch/generative/layers/feed_forward.py
```diff
@@ -93,3 +93,38 @@ def forward(self, x):
       torch.Tensor: output tensor after feedforward.
     """
     return self.w2(self.act(self.w1(x)) * self.w3(x))
+
+
+class SharedGatedFeedForward(nn.Module):
+  """Shared Gated Feedforward with customizable activation."""
+
```
haozha111, line 100: is there a paper/github reference for this feedforward function we can put here?

ai_edge_torch/generative/utilities/loader.py
yichunk: Done

ai_edge_torch/generative/layers/attention.py
yichunk: Do you mean something like this? keep the split operation into a single.

```python
if self.config.qkv_transpose_before_split:
  qkv = qkv.view(
      B, T, total_qkv, self.config.num_query_groups, self.head_dim
  )  # (B, T, total_qkv, num_query_groups, head_dim)
  qkv_axis = -3
else:
  qkv = qkv.view(
      B, T, self.config.num_query_groups, total_qkv, self.head_dim
  )  # (B, T, num_query_groups, total_qkv, head_dim)
  qkv_axis = -2
  
# Split batched computation into three.
q, k, v = qkv.split((q_per_kv, 1, 1), dim=qkv_axis)
q = q.reshape(B, T, -1, self.head_dim)
k = k.reshape(B, T, -1, self.head_dim)
v = v.reshape(B, T, -1, self.head_dim)
```

ai_edge_torch/generative/layers/builder.py
yichunk: Added

ai_edge_torch/generative/layers/feed_forward.py
yichunk: It turns out the implementation was wrong.
`QUICK_GELU` should be `QUICK_GELU(x) = x * F.sigmoid(1.702 * x)`, so it falls into `SEQUENTIAL` type.

Removed `SharedGatedFeedForward`.

haozha111: can you remove this class then?

yichunk: removed

ai_edge_torch/generative/layers/attention.py
haozha111: discussed offline, closing for now.

haozha111 (dismissed):

yichunk:
@haozha111 could you re-approve the PR? I added another commit for formatting.

haozha111 (approved):

PR #32, opened 2024-06-05
Enable test for model conversion.

haozha111:
Enable test for model conversion of GenAI examples.

We skip running the model inference since the custom op support is not done yet.

BUG=none


ai_edge_torch/generative/test/test_model_conversion.py
```diff
@@ -122,35 +127,37 @@ def test_tiny_llama_multisig(self):
...
-            copied_model,
-            (decode_token, decode_input_pos),
-            signature_name="decode",
-            num_valid_inputs=1,
-        )
-    )
+    # TODO(b/338288901): re-enable test to check output tensors.
+    skip_output_check = True
+    if skip_output_check is False:
+      # For the pytorch model, the KV cache is a persistent state internal to the model, and it
```
hheydary, line 133: Is this in-line comment still valid after the KV-Cache custom op?

hheydary (dismissed):

ai_edge_torch/generative/test/test_model_conversion.py
haozha111: cleaned up.

haozha111:
there are some tests that i need to disable to save resources, otherwise Github test will always fail/cancel.

hheydary (approved):

PR #34, opened 2024-06-05
Rename the gelu composite to avoid conflict

majiddadashi:
The gelu composite name was conflicting with a composite made in the tflite converter with the same name.

BUG=343777419


majiddadashi:
After further investigation, we learned that composite name conflict is not the source of this issue.

PR #36, opened 2024-06-06
External kv experiment

hheydary:
Experimental. Do not merge.

### DIFF
--- a/ai_edge_torch/generative/layers/experimental/kv_cache.py
+++ b/ai_edge_torch/generative/layers/experimental/kv_cache.py
@@ -14,42 +14,86 @@
 # ==============================================================================
 # Cache management utilities.
 
-import abc
-from dataclasses import dataclass
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Tuple
 
 import torch
-from torch import nn
 
 
-@dataclass
-class KVCaches:
-  """A utility dataclass for holding Key and Value cache per layer."""
+class KVCache:
+  """A utility clss for holding Key and Value caches per layer."""
 
-  k_caches: List[torch.Tensor]
-  v_caches: List[torch.Tensor]
+  def __init__(
+      self,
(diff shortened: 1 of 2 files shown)


ai_edge_torch/generative/layers/experimental/kv_cache.py
```diff
@@ -0,0 +1,99 @@
...
+# limitations under the License.
+# ==============================================================================
+# Cache management utilities.
+
+from typing import Any, Dict, List, Tuple
+
+import torch
+
+
+class KVCache:
```
haozha111, line 22: mark this as a dataclass?

ai_edge_torch/generative/layers/experimental/kv_cache.py
```diff
@@ -0,0 +1,99 @@
...
+  """A utility clss for holding Key and Value caches per layer."""
```
haozha111, line 23: class

ai_edge_torch/generative/layers/experimental/kv_cache.py
```diff
@@ -0,0 +1,99 @@
...
+
+  def __init__(
+      self,
+      num_layers: int,
+      shape: Tuple[int],
+      dtype: torch.dtype = torch.float32,
+      device: torch.device = None,
```
haozha111, line 30: is this useful for our use case?

ai_edge_torch/generative/layers/experimental/kv_cache.py
hheydary: Done

ai_edge_torch/generative/layers/experimental/kv_cache.py
hheydary: It's generally good practice to allocate on the correct device instead of moving later. Beside, it has a default and should not matter if not needed.

PR #42, opened 2024-06-09
Add GeGLU activation and change activation type to activation config

yichunk:
Add GeGLU activation and change activation type to activation config for model_config.

GeGLU activation is used in diffusion model.
And GeGLU has parameters: https://paperswithcode.com/method/geglu
Change activation type to activation config in all model_configs class.

BUG=b/311216181


haozha111 (approved):

haozha111:
nice cleanup.

PR #49, opened 2024-06-11
Add cross attention layer and refactor t5_attention

yichunk:
Add cross attention layer and refactor t5_attention to inherit from cross attention.

Test the t5 conversion and work.

BUG=b/311216181


haozha111 (approved):
nice refactoring!

PR #54, opened 2024-06-13
Update gemma and example model to use external KV.

hheydary:
Experimental external KV. Do not merge.


ai_edge_torch/generative/layers/experimental/kv_cache.py
```diff
@@ -0,0 +1,99 @@
...
+class KVCache:
+  """A utility class for holding Key and Value caches per layer."""
+
+  k_caches: List[torch.Tensor]
+  v_caches: List[torch.Tensor]
+
+  def __init__(
+      self,
+      num_layers: int,
+      shape: Tuple[int],
```
haozha111, line 33: i guess eventually we might want a list of shapes or something, so each layer's kv shape can be different.

PR #57, opened 2024-06-14
Remove 5D tensor reshape in attention layer implementation.

yichunk:
Remove 5D tensor reshape in attention layer implementation.

test on TinyLlama result, which matchs the result before removing. 

BUG=b/347108294

### DIFF
--- a/ai_edge_torch/generative/layers/attention.py
+++ b/ai_edge_torch/generative/layers/attention.py
@@ -189,20 +189,21 @@ class CausalSelfAttention(nn.Module):
 
     # Assemble into a number of query groups to support MHA, MQA and GQA.
     q_per_kv = self.config.num_heads // self.config.num_query_groups
-    total_qkv = q_per_kv + 2  # Each group has >=1 queries, 1 key, and 1 value.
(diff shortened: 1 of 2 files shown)


haozha111:
Great, have you tested if it works for the other models (phi2 and gemma)? we need to check if this change is correct for all types of attentions (MHA, MQA, GQA). Thanks!

yichunk:
> Great, have you tested if it works for the other models (phi2 and gemma)? we need to check if this change is correct for all types of attentions (MHA, MQA, GQA). Thanks!

TinyLlama is GQA, so it should cover for MHA and MQA. Let me test on phi2 and gemma to confirm.

yichunk:
> > Great, have you tested if it works for the other models (phi2 and gemma)? we need to check if this change is correct for all types of attentions (MHA, MQA, GQA). Thanks!
> 
> TinyLlama is GQA, so it should cover for MHA and MQA. Let me test on phi2 and gemma to confirm.

Test on phi2 and Gemma, and the results match the ones before.

haozha111 (approved):

haozha111 (approved):

PR #61, opened 2024-06-18
Wrap embedding tables with HLFB

paulinesho:
BUG=b/332759630

### DIFF
--- a/ai_edge_torch/convert/fx_passes/build_aten_composite_pass.py
+++ b/ai_edge_torch/convert/fx_passes/build_aten_composite_pass.py
@@ -221,12 +221,19 @@ def _aten_embedding(gm: GraphModule, node: Node):
   def embedding(*args, **kwargs):
     nonlocal op, args_mapper
     full_kwargs = args_mapper.get_full_kwargs(args, kwargs)
+    _, embedding_dim = full_kwargs['weight'].size()
+    idx = full_kwargs["indices"]
+    B, T = idx.size()
+    idx = torch.reshape(idx, (B*T,))
+
     builder = StableHLOCompositeBuilder("odml.embedding_lookup")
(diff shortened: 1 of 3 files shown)


haozha111 (approved):

ai_edge_torch/convert/fx_passes/build_aten_composite_pass.py
```diff
@@ -213,6 +213,35 @@ def is_valid_padding(padding: list[int]):
...
+
+  def embedding(*args, **kwargs):
+    nonlocal op, args_mapper
+    full_kwargs = args_mapper.get_full_kwargs(args, kwargs)
+    _, embedding_dim = full_kwargs["weight"].size()
+    idx = full_kwargs["indices"]
+    idx = idx.type(torch.int)
+    B, T = idx.size()
+
+    idx = torch.reshape(idx, (B * T,))
```
majiddadashi, line 229: I think injecting operations as we are building a composite op is adding to the places where we are modifying the program. This breaks the concept that composites are only a way to annotate the program with high level information.

I think the right design is to annotate the graph with pytorch info (build an `aten.embedding` composite) in the pytorch layer and then convert it to what tflite can consume in the tflite converter. This PR is using the tflite kernel/converter knowledge in the pytorch side of the stack.

majiddadashi (approved):

ai_edge_torch/convert/fx_passes/build_aten_composite_pass.py
haozha111: the issue here is that we want to avoid adding `reshape`s to the pytorch layer. doing this as fx passes will hide the complexity from the user.

haozha111 (approved):

PR #62, opened 2024-06-18
Add fx pass to remove sdpa composite zero mask input

chunnienc:
BUG=https://b.corp.google.com/issues/347110617

### DIFF
--- a/ai_edge_torch/convert/conversion.py
+++ b/ai_edge_torch/convert/conversion.py
@@ -30,6 +30,7 @@ from ai_edge_torch.convert.fx_passes import CanonicalizePass
 from ai_edge_torch.convert.fx_passes import InjectMlirDebuginfoPass
 from ai_edge_torch.convert.fx_passes import OptimizeLayoutTransposesPass
 from ai_edge_torch.convert.fx_passes import run_passes
+from ai_edge_torch.generative.fx_passes import run_generative_passes
 from ai_edge_torch.quantize import quant_config as qcfg
 
 os.environ["EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM"] = "1"
@@ -38,6 +39,7 @@ os.environ["EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM"] = "1"
 def _run_convert_passes(
     exported_program: ExportedProgram,
 ) -> ExportedProgram:
+  exported_program = run_generative_passes(exported_program)
   return run_passes(
       exported_program,
       [
--- a/ai_edge_torch/generative/fx_passes/__init__.py
+++ b/ai_edge_torch/generative/fx_passes/__init__.py
@@ -1,3 +1,17 @@
+# Copyright 2024 The AI Edge Torch Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
 import torch
 
 from ai_edge_torch.convert.fx_passes import CanonicalizePass
--- a/ai_edge_torch/generative/fx_passes/__init__.py
+++ b/ai_edge_torch/generative/fx_passes/__init__.py
@@ -1,0 +1,17 @@
+import torch
+
+from ai_edge_torch.convert.fx_passes import CanonicalizePass
+from ai_edge_torch.convert.fx_passes import run_passes
+from ai_edge_torch.generative.fx_passes.remove_sdpa_zero_mask_pass import RemoveSDPACompositeZeroMaskPass  # NOQA
+
+
+def run_generative_passes(
+    exported_program: torch.export.ExportedProgram,
+) -> torch.export.ExportedProgram:
+  return run_passes(
+      exported_program,
+      [
+          RemoveSDPACompositeZeroMaskPass(),
+          CanonicalizePass(),
+      ],
+  )
--- a/ai_edge_torch/generative/fx_passes/test/test_remove_sdpa_zero_mask_pass.py
+++ b/ai_edge_torch/generative/fx_passes/test/test_remove_sdpa_zero_mask_pass.py
@@ -1,0 +1,125 @@
+# Copyright 2024 The AI Edge Torch Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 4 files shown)


ai_edge_torch/convert/conversion.py
```diff
@@ -38,6 +39,7 @@
 def _run_convert_passes(
     exported_program: ExportedProgram,
 ) -> ExportedProgram:
+  exported_program = run_generative_passes(exported_program)
```
majiddadashi, line 42: generative API is just a user of ai-edge-torch. Should we 
1. let the user inject passes into our conversion pipeline OR
2. ask them to first apply their pass to their ExportedProgram and then pass in the ExportedProgram?

haozha111: ideally this pass should just run for generative api model export flow (running by default).

chunnienc: I think both won't work simply:
1. Our pass system is hidden from users now, and I don't expect users to understand this either.
2. Our API `ai_edge_torch.convert` hides the exported program details from users

For this PR I can confirm that the pass won't interact with all the other conversion passes, for other passes in the future we should review case-by-case to make sure no unexpected interactions.

A better way is to enable the generative pass pipeline when we detect the model is composed of generative layers, while this identifier does not exist today. We can come back to this when there are more requests about generative passes come in later.

haozha111: sgtm. the current odml.sdpa won't appear in any non-gen ai models, it should be fine to just run it in the default conversion flow. However we should clean this up later.

haozha111 (approved):

ai_edge_torch/convert/conversion.py
majiddadashi: Adding user specific passes to `ai_edge_torch.convert` is cluttering the main conversion code. Can this be created the way it is expected in the generative layer?

chunnienc: cc @arfaian

haozha111: unless you have something like `ai_edge_torch.converter.apply_gpu_passes` in the gen ai conversion flow?

haozha111: maybe this is a nicer solution:

is it possible to associate a pattern/function (https://github.com/google-ai-edge/ai-edge-torch/blob/1a4fe129f98a391c2d96fdb3f3f4d49aa4810562/ai_edge_torch/generative/layers/scaled_dot_product_attention.py#L69) with a fx rewriter rule? If that could be done, then we don't need to introduce the clutter mentioned here, instead the fx rewriter will automatically apply rule 'X' when seeing pattern 'Y'.

arfaian: For now adding these passes underneath the hood and leaving the `ai_edge_torch.convert` API untouched is the way to go (assuming the passes are truly a no-op when the layers are not present as mentioned). Majid's right in that it will add some clutter for the time being but I think that's acceptable to buy us some time to bake a better solution. Long-term we will want to think of a user journey where 3p users may want to apply a similar approach for a model (custom layers and custom passes) and how that will look but this will need more thought and discussion (e.g. are we only going to allow fx passes? what about mlir passes and what's the story going to be on the JAX side for equivalent functionality?).

chunnienc: > maybe this is a nicer solution:
> 
> is it possible to associate a pattern/function (
> 
> https://github.com/google-ai-edge/ai-edge-torch/blob/1a4fe129f98a391c2d96fdb3f3f4d49aa4810562/ai_edge_torch/generative/layers/scaled_dot_product_attention.py#L69
> 
> ) with a fx rewriter rule? If that could be done, then we don't need to introduce the clutter mentioned here, instead the fx rewriter will automatically apply rule 'X' when seeing pattern 'Y'.

No. If this idea works for you I'd recommend to change the composite/layer authoring with what Yi-Chun supposed to do (don't mark `mask` as composite input if the input is None, and don't feed torch.zeros anywhere).

and thanks @arfaian for the reply. If this PR looks good to all you guys I'll just merge it as is, and come back later if needed.

majiddadashi: curious why this logic can't be added where the composite is [being built](https://github.com/google-ai-edge/ai-edge-torch/blob/1a4fe129f98a391c2d96fdb3f3f4d49aa4810562/ai_edge_torch/generative/layers/scaled_dot_product_attention.py#L96)? I mean not passing `mask` through `mark_inputs` if it's all zeros?

chunnienc: torch.export.export does not work if the model depends on tensor value, unless you expect it to export it to an aten.where, which is not the case here.

The way to fix in authoring is to not pass torch.zeros at all and leave the input to be a None. While I guess @haozha111 has some concerns on that so he is looking for solution with fx pass.

majiddadashi: how about `torch.cond`?

```
def with_mark(x):
   return builder.mark_inputs(x)

def without_mark(x):
   return x

mask = torch.cond(torch.all(mask == 0), with_mark, without_mark, (mask,))
```

chunnienc: > how about `torch.cond`?
> 
> ```
> def with_mark(x):
>    return builder.mark_inputs(x)
> 
> def without_mark(x):
>    return x
> 
> mask = torch.cond(torch.all(mask == 0), with_mark, without_mark, (mask,))
> ```

This will break the composite builder. Also, I guess in general the tfl.select is more unexpected.

haozha111: > torch.export.export does not work if the model depends on tensor value, unless you expect it to export it to an aten.where, which is not the case here.
> 
> The way to fix in authoring is to not pass torch.zeros at all and leave the input to be a None. While I guess @haozha111 has some concerns on that so he is looking for solution with fx pass.

yes, a fx pass fix is preferred since it doesn't need any changes in the authoring layer.

ai_edge_torch/generative/fx_passes/test/test_remove_sdpa_zero_mask_pass.py
arfaian, line 1: Naming convention is <filename>_test.py, right? At least internally in g3. Any reason it's a prefix here?

ai_edge_torch/generative/fx_passes/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def run_generative_passes(
+    exported_program: torch.export.ExportedProgram,
+) -> torch.export.ExportedProgram:
+  return run_passes(
+      exported_program,
+      [
+          RemoveSDPACompositeZeroMaskPass(),
+          CanonicalizePass(),
```
arfaian, line 29: For my own understanding, why do you need to canonicalize again?

ai_edge_torch/generative/fx_passes/test/test_remove_sdpa_zero_mask_pass.py
chunnienc: The test file naming in ai_edge_torch is prefix `test_`. This is the requirement for test file to be detected by ` python -m unittest discover`. We keep this convention even if we've migrated to pytest now.

ai_edge_torch/generative/fx_passes/__init__.py
chunnienc: To remove the node with identity python function. In general our fx passes expect the fx graph to be a graph of aten ops, while the fx graph node target can be any python function. The CanonicalizePass can be considered as a dynamo retracing here to turn whatever python fx graph into the aten ops only fx graph.

arfaian (approved):

PR #78, opened 2024-07-02
[Prototype] Align SDPA definition with the gpu accelerator

Linchenn:


Linchenn:
cc @haozha111 @advaitjain 

With this pass, the converted tflite could be run on the GPU accelerator (though it has a tiny bug to be fixed). 

If we finally decide to use this approach, we additionally need to figure out how to diverge conversion between the CPU accelerator and the GPU accelerator.

haozha111:
> cc @haozha111 @advaitjain
> 
> With this pass, the converted tflite could be run on the GPU accelerator (though it has a tiny bug to be fixed).
> 
> If we finally decide to use this approach, we additionally need to figure out how to diverge conversion between the CPU accelerator and the GPU accelerator.

thanks! can we make this PR in draft mode? we don't need any code reviews for it now.

PR #87, opened 2024-07-11
Allow explictly setting head_dim instead of deducing it.

mbrenon:
This is a requirement for the upcoming OpenELM models.

Note that head_dim is also moved down to the attention config: this parameter is related to attention so it has no reason to live under the main model config. 

Please review carefully to make sure I'm not breaking anything here :)

BUG=https://b.corp.google.com/issues/352478939

### DIFF
--- a/ai_edge_torch/generative/examples/stable_diffusion/diffusion.py
+++ b/ai_edge_torch/generative/examples/stable_diffusion/diffusion.py
@@ -250,6 +250,7 @@ class Diffusion(nn.Module):
 
     attention_config = layers_cfg.AttentionConfig(
         num_heads=config.transformer_num_attention_heads,
+        head_dim=block_out_channels[0] // config.transformer_num_attention_heads,
         num_query_groups=config.transformer_num_attention_heads,
         rotary_percentage=0.0,
         qkv_transpose_before_split=True,
--- a/ai_edge_torch/generative/layers/attention.py
+++ b/ai_edge_torch/generative/layers/attention.py
@@ -136,14 +136,16 @@ class CausalSelfAttention(nn.Module):
       enable_hlfb (bool): whether hlfb is enabled or not.
     """
     super().__init__()
-    self.head_dim = dim // config.num_heads
-    shape = (config.num_heads + 2 * config.num_query_groups) * self.head_dim
-    # Key, query, value projections for all heads.
-    self.qkv_projection = nn.Linear(dim, shape, bias=config.qkv_use_bias)
-    self.output_projection = nn.Linear(dim, dim, bias=config.output_proj_use_bias)
     self.config = config
     self.kv_cache = None
     self.batch_size = batch_size
+    qkv_shape = (config.num_heads + 2 * config.num_query_groups) * config.head_dim
+    output_shape = config.num_heads * config.head_dim
+    # Key, query, value projections for all heads.
+    self.qkv_projection = nn.Linear(dim, qkv_shape, bias=config.qkv_use_bias)
+    self.output_projection = nn.Linear(
+        output_shape, dim, bias=config.output_proj_use_bias
+    )
 
     # Build a k/v cache with size (batch_size, kv_cache_max, n_heads, head_dim).
     if config.enable_kv_cache:
@@ -151,7 +153,7 @@ class CausalSelfAttention(nn.Module):
           batch_size,
           kv_cache_max,
           config.num_query_groups,
-          self.head_dim,
+          config.head_dim,
           enable_hlfb,
       )
 
@@ -180,7 +182,7 @@ class CausalSelfAttention(nn.Module):
       output activation from this self attention layer.
     """
     # Batch size, sequence length, embedding dimensionality.
-    B, T, E = x.size()
+    B, T, _ = x.size()
     assert (
         B == self.batch_size
     ), "batch size of input tensor must match with the batch size specified in the model configuration."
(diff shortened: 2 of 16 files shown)


ai_edge_torch/generative/layers/model_config.py
```diff
@@ -55,9 +55,10 @@ class FeedForwardType(enum.Enum):
 
 @dataclass
 class AttentionConfig:
-  """Attention moduel's parameters."""
+  """Attention model's parameters."""
```
talumbau, line 58: I think `module` is probably the best choice and follows PyTorch naming convention here. Both are better than the existing `moduel` though :-)

haozha111: +1

ai_edge_torch/generative/layers/model_config.py
```diff
@@ -55,9 +55,10 @@ class FeedForwardType(enum.Enum):
...
 
   num_heads: int
+  head_dim: int
```
haozha111, line 61: i guess in most common cases (other than OpenELM) the `head_dim` could be automatically derived as the current library does, this will eliminate setting one more parameter when constructing the model.

Can we keep the current behavior (so that most models don't need a change and supply this additional parameter), but allow the parameter to be overwritten, or derived in another user supplied lamda?

ai_edge_torch/generative/layers/attention.py
```diff
@@ -136,22 +136,24 @@ def __init__(
...
     self.config = config
     self.kv_cache = None
     self.batch_size = batch_size
+    qkv_shape = (config.num_heads + 2 * config.num_query_groups) * config.head_dim
+    output_shape = config.num_heads * config.head_dim
+    # Key, query, value projections for all heads.
+    self.qkv_projection = nn.Linear(dim, qkv_shape, bias=config.qkv_use_bias)
+    self.output_projection = nn.Linear(
+        output_shape, dim, bias=config.output_proj_use_bias
+    )
```
talumbau, line 148: Yes. This change is also needed for Gemma v2. Thanks @mbrenon! My plan will be to base my branch off this one.

ai_edge_torch/generative/layers/model_config.py
haozha111: this is kinda important design decision if we want to keep the `ModelConfig`api easy and simple to use, without exposing too many knobs to the user.

haozha111: @talumbau @yichunk @hheydary any thoughts you could share here?

ai_edge_torch/generative/utilities/stable_diffusion_loader.py
```diff
@@ -698,6 +698,7 @@ def load(
 
     attention_config = layers_config.AttentionConfig(
         num_heads=config.transformer_num_attention_heads,
+        head_dim=config.block_out_channels[0] // config.transformer_num_attention_heads,
```
talumbau, line 701: This change will need @yichunk to review

ai_edge_torch/generative/layers/model_config.py
talumbau: GemmaV2 needs this exact change and I was about to make the changes to CausalSelfAttention that Maxime does here. I suspect more architectures will use this trick to squeeze down the parameter count a bit. So, I am +1 to being explicit here.

haozha111: cool, sgtm. If this is going to be common, exposing it explicitly sounds like a good plan.

ai_edge_torch/generative/layers/attention.py
mbrenon: Great!

ai_edge_torch/generative/utilities/stable_diffusion_loader.py
mbrenon: For context: I'm just using the way head_dim was previously defined [here](https://github.com/google-ai-edge/ai-edge-torch/blob/9fa04010af0aabfb1b97da1675202432155d138b/ai_edge_torch/generative/layers/model_config.py#L157-L158)

talumbau:
I ran inference on the example decoder models with this branch and everything looked good. Also converted tiny_llama with no issues. Would be great to get an official thumbs up from someone who has worked with the Stable Diffusion model.

talumbau (approved):

haozha111:
@yichunk could you help review the changes to SD code? thanks!

talumbau:
Actually, I see we need another change for the T5 encoder/decoder models, similar to the other decoder-only models. Please add

```
 head_dim=64
```

To the `AttentionConfig` in get_model_config_t5() function in `t5.py`. Thanks @mbrenon!

haozha111:
@yichunk could you help approve this PR? thanks!

ai_edge_torch/generative/examples/stable_diffusion/diffusion.py
```diff
@@ -250,6 +250,7 @@ def __init__(self, config: unet_cfg.DiffusionModelConfig):
 
     attention_config = layers_cfg.AttentionConfig(
         num_heads=config.transformer_num_attention_heads,
+        head_dim=block_out_channels[0] // config.transformer_num_attention_heads,
```
yichunk, line 253: Attention dims are different for each block. Should construct attention config in the down_encoders loop below. Check `AttentionBlock2DConfig` and `CrossAttentionBlock2DConfig`.

Also, the attention configs in mid_block and up_decoders should be fixed as well.

ai_edge_torch/generative/utilities/stable_diffusion_loader.py
yichunk: Same as config defined in `diffusion.py`. The head dim will be different for each block.

talumbau:
Closing in favor of #120

PR #89, opened 2024-07-15
Add support for composite kwargs

hheydary:
Add support for composite keyword arguments.

BUG=b/353293690

### DIFF
--- a/ai_edge_torch/convert/conversion_utils.py
+++ b/ai_edge_torch/convert/conversion_utils.py
@@ -80,24 +80,43 @@ class Signature:
       names.append(f"args_{i}")
 
     kwargs_names = self._flat_kwarg_names(
-        kwargs_spec.children_specs, [kwargs_spec.context]
+        kwargs_spec.children_specs, kwargs_spec.context
     )
     names.extend(kwargs_names)
     return names
 
   def _flat_kwarg_names(self, specs, context) -> List[str]:
-    if context is None:
-      return [f"{i}" for i in range(len(specs))] if specs else []
-
     flat_names = []
-    for prefix, spec in zip(context[0], specs):
-      leaf_flat_names = self._flat_kwarg_names(spec.children_specs, spec.context)
-      if leaf_flat_names:
-        flat_names.extend([f"{prefix}_{name}" for name in leaf_flat_names])
-      else:
-        flat_names.append(prefix)
+    if context is None:
+      for i, spec in enumerate(specs):
+        if spec.children_specs:
+          flat_names.extend(
+              [
+                  f"{i}_{name}"
+                  for name in self._flat_kwarg_names(spec.children_specs, spec.context)
+              ]
+          )
+        else:
+          flat_names.append(f"{i}")
+    else:
+      flat_ctx = self._flatten_list(context)
+      for prefix, spec in zip(flat_ctx, specs):
+        leaf_flat_names = self._flat_kwarg_names(spec.children_specs, spec.context)
+        if leaf_flat_names:
+          flat_names.extend([f"{prefix}_{name}" for name in leaf_flat_names])
+        else:
+          flat_names.append(prefix)
 
     return flat_names
+
+  def _flatten_list(self, l: List) -> List:
+    flattened = []
+    for item in l:
+      if isinstance(item, list):
+        flattened.extend(self._flatten_list(item))
+      else:
+        flattened.append(item)
+    return flattened
 
   @property
   def flat_args(self) -> tuple[Any]:
--- a/ai_edge_torch/convert/conversion_utils.py
+++ b/ai_edge_torch/convert/conversion_utils.py
@@ -20,7 +20,7 @@ import gc
 import itertools
 import logging
 import tempfile
-from typing import Any, Dict, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import torch
 import torch.utils._pytree as pytree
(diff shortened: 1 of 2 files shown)


ai_edge_torch/convert/test/test_convert.py
```diff
@@ -306,6 +308,36 @@ def forward(self, x, y):
...
+
+    torch.export.register_dataclass(TestContainer, serialized_type_name="TestContainer")
+
+    class SampleModel(torch.nn.Module):
+
+      def forward(self, x: torch.Tensor, y: torch.Tensor, z: TestContainer):
+        return x + y + z.data_1 + z.data_2[0] + z.data_2[1]
+
+    args_gen = lambda: (torch.randn(10, 10),)
+    kwargs_gen = lambda: dict(
```
majiddadashi, line 329: why not just create the object as opposed to creating an args generator function?

ai_edge_torch/convert/test/test_convert.py
```diff
@@ -306,6 +308,36 @@ def forward(self, x, y):
...
+        y=torch.randn(10, 10),
+        z=TestContainer(
+            data_1=torch.randn(10, 10),
+            data_2=(torch.randn(10, 10), torch.randn(10, 10)),
+        ),
+    )
+
+    model = SampleModel().eval()
+    edge_model = ai_edge_torch.convert(model, args_gen(), kwargs_gen())
+    self.assertIsNotNone(edge_model)
```
majiddadashi, line 339: Please extend this to run the model and make sure the result is correct.

ai_edge_torch/convert/test/test_convert.py
```diff
@@ -306,6 +308,36 @@ def forward(self, x, y):
...
+    """
+    Test converting a simple model with both sample_args and sample_kwargs.
+    """
+
+    @dataclass
+    class TestContainer:
+      data_1: torch.Tensor
+      data_2: Tuple[torch.Tensor, torch.Tensor]
+
+    torch.export.register_dataclass(TestContainer, serialized_type_name="TestContainer")
```
majiddadashi, line 321: in the current state, wondering if we even need this line given that we flatten the args before exporting?

ai_edge_torch/convert/conversion_utils.py
```diff
@@ -79,28 +79,30 @@ def flat_arg_names(self) -> list[str]:
     for i in range(args_spec.num_leaves):
       names.append(f"args_{i}")
 
-    dict_context = (
-        kwargs_spec.context
```
chunnienc, line 83: please respect the if-else branch and special case for defaultdict in your new impl, thanks

ai_edge_torch/convert/conversion_utils.py
```diff
@@ -79,28 +79,30 @@ def flat_arg_names(self) -> list[str]:
...
+        flat_names.append(prefix)
+
+    return flat_names
 
   @property
-  def flat_args(self) -> tuple[torch.Tensor]:
-    return tuple(pytree.tree_flatten(self._normalized_sample_args_kwargs)[0])
+  def flat_args(self) -> tuple[Any]:
+    args, kwargs = self._normalized_sample_args_kwargs
+    return tuple([*args, *kwargs.values()])
```
chunnienc, line 105: Please keep the original impl which returns the flat tensor arg list. The intent is to make `len(flat_args) == len(flat_arg_names)` as a way to debug when there is an error, and implicitly disable internal flattening in torch.export.export as that may have a conflict to this external flattening.

ai_edge_torch/convert/test/test_convert.py
```diff
@@ -306,6 +308,36 @@ def forward(self, x, y):
         model_coverage.compare_tflite_torch(edge_model, model, args_gen, kwargs_gen)
     )
 
+  def test_convert_model_with_args_nested_kwargs(self):
+    """
+    Test converting a simple model with both sample_args and sample_kwargs.
```
chunnienc, line 313: update test docstring

chunnienc (changes requested):

ai_edge_torch/convert/test/test_convert.py
hheydary: Done

ai_edge_torch/convert/conversion_utils.py
hheydary: I am flattening this context in the updated implementation that fixes the bugs. PTAL.

ai_edge_torch/convert/test/test_convert.py
hheydary: Done

ai_edge_torch/convert/conversion_utils.py
hheydary: The problem is that during conversion, we use sig.flat_args.
```
  exported_programs: torch.export.ExportedProgram = [
      torch.export.export(sig.module, sig.flat_args, dynamic_shapes=sig.dynamic_shapes)
      for sig in signatures
  ]
  ```
However, with nested arguments, the existing logic would give you the wrong number of keyword arguments.

Please note that on authoring side (i.e, pytorch), the structures are maintained and not flattened.

Comment (chunnienc) [reply ai_edge_torch/convert/conversion_utils.py:124]: > the existing logic would give you the wrong number of keyword arguments.

Could you explain more on this?

> Please note that on authoring side (i.e, pytorch), the structures are maintained and not flattened.

While the flattening would always happen before the conversion. The intent is to make logics here the only source of truth for the flattening rule. The worst case is we deep flatten the kwargs here to get the flat name list and pass the dict values as is to  torch.export.export, but then torch export uses another rule/order to flatten the dict.

Comment (hheydary) [reply ai_edge_torch/convert/conversion_utils.py:124]: Let's take this example:
```
    class SampleModel1(torch.nn.Module):

      def forward(self, x: torch.Tensor, y: torch.Tensor, z: TestContainer):
        return x + y + z.data_1 + z.data_2[0] + z.data_2[1]

    args = (torch.randn(10, 10),)
    kwargs = dict(
        y=torch.randn(10, 10),
        z=TestContainer(
            data_1=torch.randn(10, 10),
            data_2=(torch.randn(10, 10), torch.randn(10, 10)),
        ),
    )
```
On pytorch side, the model would expect **three** arguments (x, y, z). However, the following logic:
```
return tuple(pytree.tree_flatten(self._normalized_sample_args_kwargs)[0])
```
would return **five** values (the flattened args). This triggers an error in `torch.export.export`.

Comment (chunnienc) [reply ai_edge_torch/convert/conversion_utils.py:124]: oh I see, yes you're right on this. Thanks for the explanation.

We may still run into issues with this impl when the model.forward expects kw-only args. But I guess there are no better way to solve this right now.

File: ai_edge_torch/convert/test/test_convert.py [hunk 7]
```diff
@@ -306,6 +310,99 @@ def forward(self, x, y):
...
+    flat_inputs = {
+        "args_0": args[0].numpy(),
+        "y": kwargs["y"].numpy(),
+        "z_data_1": kwargs["z"].data_1.numpy(),
+        "z_data_2_0": kwargs["z"].data_2[0].numpy(),
+        "z_data_2_1": kwargs["z"].data_2[1].numpy(),
+    }
+    self._args_kwargs_test_helper(SampleModel1(), args, kwargs, flat_inputs)
+
+    #### case 2
```
chunnienc, line 348: pls separate this into multiple tests. Name them `test_convert_model_with_args_nested_kwargs_{n}` should be good enough.

ai_edge_torch/convert/test/test_convert.py
```diff
@@ -306,6 +310,99 @@ def forward(self, x, y):
...
+    flat_inputs = {
+        "args_0": args[0].numpy(),
+        "y": kwargs["y"].numpy(),
+        "z_data_1": kwargs["z"].data_1.numpy(),
+        "z_data_2_0_foo": kwargs["z"].data_2[0]["foo"].numpy(),
+        "z_data_2_1": kwargs["z"].data_2[1].numpy(),
+    }
+    self._args_kwargs_test_helper(SampleModel3(), args, kwargs, flat_inputs)
+
+  def _args_kwargs_test_helper(self, model, args, kwargs, flat_inputs):
```
chunnienc, line 394: more meaningful name like `compare_tflite_torch_args_kwargs`

chunnienc (approved):

ai_edge_torch/convert/test/test_convert.py
majiddadashi: I think this was marked as resolved by mistake (?)

majiddadashi (changes requested):
Please extend this to allow running the model with the arguments the model can be traced with.

ai_edge_torch/convert/test/test_convert.py
hheydary: Done.

majiddadashi (approved):
Approving this PR and @hheydary will send a subsequent PR to modify model.py to make execution consistent.

PR #90, opened 2024-07-15
External kv

hheydary:
Add experimental support for external KV cache.

BUG=b/353293690


ai_edge_torch/generative/examples/experimental/gemma/convert_to_tflite.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+import os
+from pathlib import Path
+
+import torch
+
+import ai_edge_torch
+from ai_edge_torch.generative.examples.experimental.gemma import gemma
+from ai_edge_torch.generative.layers.experimental import kv_cache as kv_utils
```
talumbau, line 27: Just for extreme clarity for now, I think the module name should indicate "external" in some way, e.g. `ekv_cache`, `ext_kv_cache`, something like that. We can do a global search/replace to just a standard "KV Cache" name when it comes out of experimental.

ai_edge_torch/generative/layers/experimental/kv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+        config.head_dim,
+    )
+    k = torch.zeros(shape, dtype=dtype, device=device)
+    v = torch.zeros(shape, dtype=dtype, device=device)
+    obj = cls(k_cache=k, v_cache=v)
+    return obj
+
+
+@dataclass
+class KVCache:
```
talumbau, line 58: Would use name `ExpKVCache` or `EKVCache`. Something like that. It seems like a simple way to avoid confusion (or even name stomping) until we are ready to make the switchover.

ai_edge_torch/generative/layers/experimental/kv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+  for i, kv_entry in enumerate(kvc.caches):
+    flattened.append(kv_entry.k_cache)
+    flat_names.append(f"k_{i}")
+    flattened.append(kv_entry.v_cache)
+    flat_names.append(f"v_{i}")
+  return flattened, [flat_names, none_names]
+
+
+def _flatten_kvc_with_keys(kvc: KVCache) -> Tuple[List, List]:
+  flattened, (flat_names, none_names) = _flatten_kvc(kvc)
```
talumbau, line 101: At first I thought I understood this, but now I realize I don't. Why is `none_names` created a returned? The function implementation doesn't use it and the caller who gets it back doesn't use it.

hheydary: It's the pytorch convention. They define `context` as list of names and non-names.

ai_edge_torch/generative/examples/experimental/gemma/convert_to_tflite.py
hheydary: Good point. Updated.

ai_edge_torch/generative/layers/experimental/kv_cache.py
hheydary: Done.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+from ai_edge_torch import hlfb
+from ai_edge_torch.generative.layers import model_config
+
+
+@dataclass
+class KVCacheEntry:
+  """A single cache entry include K and V caches.
+
+  Returns:
+      _type_: _description_
```
haozha111, line 32: this comment needs to be updated.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+
+  @classmethod
+  def from_model_config(
+      cls,
+      config: model_config.ModelConfig,
+      dtype: torch.dtype = torch.float32,
+      device: torch.device = None,
+  ) -> "KVCacheEntry":
+    shape = (
+        1,
```
haozha111, line 46: should this be batch_size?

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+        config.head_dim,
+    )
+    k = torch.zeros(shape, dtype=dtype, device=device)
+    v = torch.zeros(shape, dtype=dtype, device=device)
+    obj = cls(k_cache=k, v_cache=v)
+    return obj
+
+
+@dataclass
+class EKVCache:
```
haozha111, line 58: ExternalKVCache?

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    caches = [
+        KVCacheEntry.from_model_config(config, dtype, device)
+        for _ in range(config.num_layers)
+    ]
+    obj = cls(caches=tuple(caches))
+    return obj
+
+
+def _flatten_kvc(kvc: EKVCache) -> Tuple[List, List]:
```
haozha111, line 88: can we add some unit tests for functions in this file?

haozha111: this will also help readers know what those functions are doing by looking at the tests

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+  return flattened, [flat_names, none_names]
+
+
+def _flatten_kvc_with_keys(kvc: EKVCache) -> Tuple[List, List]:
+  flattened, (flat_names, none_names) = _flatten_kvc(kvc)
+  return [(pytree.MappingKey(k), v) for k, v in zip(flat_names, flattened)], flat_names
+
+
+def _unflatten_kvc(values: List[torch.Tensor], context: Tuple[List, List]) -> EKVCache:
+  assert len(values) % 2 == 0
```
haozha111, line 106: can you also output an error message here

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def update(
+    cache: KVCacheEntry,
+    input_pos: torch.Tensor,
+    k_slice: torch.Tensor,
+    v_slice: torch.Tensor,
+    use_hlfb: bool = True,
+) -> KVCacheEntry:
+  """Out of place update of Cach buffer.
```
haozha111, line 136: cache buffer

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+
+  Args:
+      cache (KVCacheEntry): The original cache buffer.
+      input_pos (torch.Tensor): The update slice positions.
```
haozha111, line 140: is it int32 tensor?

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+      k_slice (torch.Tensor): The K slice to be updated in the new cache.
+      v_slice (torch.Tensor): The V slice to be updated in the new cache.
+      use_hlfb (bool, optional): Whether the op is annotated for export.
```
haozha111, line 143: i think the comment should be more specific.. whether the op will be exported with high level function boundary?

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+def _update_kv_base_impl(
+    cache: KVCacheEntry,
+    input_pos: torch.Tensor,
+    k_slice: torch.Tensor,
+    v_slice: torch.Tensor,
+) -> KVCacheEntry:
+  k = cache.k_cache.index_copy(1, input_pos, k_slice)
+  v = cache.v_cache.index_copy(1, input_pos, v_slice)
+  updated_cach = KVCacheEntry(k, v)
+  return updated_cach
```
haozha111, line 162: updated_cache

ai_edge_torch/generative/layers/experimental/ekv_cache.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def _update_kv_hlfb_impl(
+    cache: KVCacheEntry,
+    input_pos: torch.Tensor,
+    k_slice: torch.Tensor,
+    v_slice: torch.Tensor,
+) -> Tuple[torch.Tensor, torch.Tensor]:
```
haozha111, line 170: -> KVCacheEntry?

ai_edge_torch/generative/layers/experimental/ekv_cache.py
haozha111: fine to use EKVCache since eventually this will be renamed.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
hheydary: It could be batch size. However, we are still limiting batch size to be 1 on the runtime side. So it might be easier to trap this here as opposed to letting the user to convert a model with larger batch size only to see that not working on the device.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
hheydary: Added some unit tests that check object serialization. However, please note that these are registered to pytorch export module so it has to go through that and these functions would not be tested directly by themselves.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
hheydary: Done.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
hheydary: Done.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
hheydary: The update uses index_copy which uses long tensor. So still int64 type.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
hheydary: Done.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
hheydary: Done.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
hheydary: Done.

ai_edge_torch/generative/layers/experimental/ekv_cache.py
hheydary: Done.

ai_edge_torch/generative/layers/experimental/attention.py
```diff
@@ -174,10 +172,9 @@ def forward(
     k = torch.cat((k_roped, k[..., n_elem:]), dim=-1)
 
     if kv_cache is not None:
-      k_cache, v_cache = kv_cache
-      k = k_cache.index_copy_(1, input_pos, k)
-      v = v_cache.index_copy_(1, input_pos, v)
-      kv_cache = k, v
+      kv_cache = kv_utils.update(kv_cache, input_pos, k, v, use_hlfb=self.enable_hlfb)
```
talumbau, line 175: I was surprised by this, in that the logic is to decide on a per-inference basis which function is called (the HLFB version or not). I would have thought that the function to call would be set as an attribute on the module at `__init__` time, so that it would just look like:

```
kv_cache = self.cache_update(kv_cache, input_pos, k, v)
```

where `self.cache_update` is set via something like:

```
self.cach_update = kv_utils._update_kv_hlfb_impl if enable_hlfb else kv_utils._update_kv_base_impl
```

If it traces and exports the same either way then ultimately it's not a big deal, just seems a bit odd to essentially do a dynamic function dispatch at forward time. You can resolve if you have considered this and discarded the idea, but curious to know what you think.

hheydary: I did think about it a bit when implementing this. Originally, I had `update` as a part of `EKVCache` object but that did not make sense since we are doing an out of place update. Hence `update` became a stand alone function within `ekv_cache` module.

The option were then to either define two variant of `update` within that modules and make one of them the class kv update function (similar to how sdpa is implemented right now) or have a single variant that hides the details. The first approach would require having an if-else in all the model classes that use kv_update, i.e., 
```
    self.sdpa_func = (
        scaled_dot_product_attention_with_hlfb
        if enable_hlfb
        else scaled_dot_product_attention
    )
```
That's why I went with a single `update` function that resolves to the correct one based on the input arguments. At the end of the day, the exported program looks the same and it's more of a matter of how clean the authoring story is.

talumbau: Happy to resolve for now if you thought it through and came to a difference conclusion. Can revisit later if needed.

PR #98, opened 2024-07-18
[Prototype / Do not submit] Add StableHLO Composite for GroupNorm

Linchenn:


ai_edge_torch/generative/layers/group_norm.py
```diff
@@ -0,0 +1,57 @@
...
+
+  B, C, H, W = y.shape
+  y = y.view(B, C, H * W)
+  y = y.transpose(-1, -2)
+  y = y.view(B, H, W, C)
+
+  y = builder.mark_outputs(y)
+
+  y = y.view(B, H * W, C)
+  y = y.transpose(-1, -2)
```
chunnienc, line 55: do you mean https://pytorch.org/docs/stable/generated/torch.permute.html ?

Linchenn: great catch, thanks!

PR #102, opened 2024-07-24
Fix missing group norm in converted TF Lite models.

yichunk:
Fix missing group norm in converted TF Lite models.

Verified the e2e pipeline with newly converted models.

BUG=b/355109848


chunnienc:
😅

chunnienc (dismissed):

chunnienc (approved):

PR #112, opened 2024-07-29
Add advanced usage section to quantization README

paulinesho:
BUG=b/356164136


ai_edge_torch/generative/quantize/README.md
```diff
@@ -18,7 +18,29 @@ Once converted, you will get a quantized `.tflite` model which will be ready for
 
 In the current release, the following schemes are supported:
 
-* Dynamic range quantization with FP32 activations and INT8 weights for linear ops
-* FP16 quantization with FP16 weights and FP32 activations and computation for all ops
+* Dynamic range quantization: FP32 activations, INT8 weights, and INT8 computation
```
talumbau, line 21: Re: `INT8 computation`, I have been wondering for some time about using a more precise term here. The alternatives that appeal to me are:

1.  "integer accumulation"

2. "int8 operands"

Either one of those would work I think. And similar for `FP32 computation` -> "floating point accumulation"/"fp32 operands". Maybe it's a small difference, but I think it does provide additional clarity on what is meant by "computation" here. What do you think?

ai_edge_torch/generative/quantize/README.md
```diff
@@ -18,7 +18,29 @@ Once converted, you will get a quantized `.tflite` model which will be ready for
...
+* Weight-only quantization: FP32 activations, INT8 weights, and FP32 computation
+* FP16 quantization: FP16 weights, FP32 activations and computation for all ops
+
+These correspond to the available recipes in `quant_recipes.py`.
+
+## Advanced usage
+
+In addition to configuring quantization using pre-configured recipes in `quant_recipes.py`. Users can also customize their recipes according to their specific needs using the `LayerQuantRecipe` and `GenerativeQuantRecipe` API in `quant_recipe.py`:
```
talumbau, line 29: .... "`quant_recipes.py`. Users can ..." ---> "`quant_recipes.py`, users can...."

ai_edge_torch/generative/quantize/README.md
```diff
@@ -18,7 +18,29 @@ Once converted, you will get a quantized `.tflite` model which will be ready for
...
+
+`LayerQuantRecipe` specifies at a Generative API layer (`ai_edge_torch/generative/layers`) level how ops within should be quantized. `GenerativeQuantRecipe` specifies at a model level how each component of a Generative API model should be quantized. With these selective quantization can be configured as follows:
```
talumbau, line 31: maybe "With these configuration classes, selective quantization...."

ai_edge_torch/generative/quantize/README.md
paulinesho: I've been interpreting "computation" as math operations or "accumulation and operands", so INT8 computation probably doesn't make sense and at the minimum we should change it to integer computation and floating point computation. "Accumulation" feels a bit too specific to certain ops IMO. WDYT of just "integer computation" or "floating point computation"?

ai_edge_torch/generative/quantize/README.md
paulinesho: Done.

ai_edge_torch/generative/quantize/README.md
paulinesho: Done.

ai_edge_torch/generative/quantize/README.md
talumbau: yes "integer" (or "float") "computation" for now sounds good.

ai_edge_torch/generative/quantize/README.md
```diff
@@ -18,7 +18,29 @@ Once converted, you will get a quantized `.tflite` model which will be ready for
...
+      generative_recipe=quant_recipe.GenerativeQuantRecipe(
+          default=create_layer_quant_fp16(),
+          embedding=create_layer_quant_int8_dynamic(),
+          attention=create_layer_quant_int8_weight_only(),
+          feedforward=create_layer_quant_int8_dynamic(),
+      )
+  )
+```
+
+For example, this recipe specifies that the embedding table, attention, and feedforward layers should be quantized to INT8. Specifically for attention layers the computation should also be in FP32. All other ops should be quantized to the default scheme which is specified as FP16.
```
talumbau, line 45: "Specifically, for attention layers the computation should be in FP32."

(added a comma and removed `also`)

talumbau (dismissed):
Please me the small change noted, but overall LGTM!

ai_edge_torch/generative/quantize/README.md
paulinesho: Done. PTAL.

talumbau (approved):

PR #120, opened 2024-08-01
Explicit head_dim configuration

talumbau:
Explicitly set head_dim in AttentionConfig

Based on @mbrenon's https://github.com/google-ai-edge/ai-edge-torch/pull/87 this PR adds head_dim configuration for

 - CrossAttention modules
 - T5 AttentionConfigs
 - StableDiffusion

BUG=https://b.corp.google.com/issues/352144652

### DIFF
--- a/ai_edge_torch/generative/utilities/stable_diffusion_loader.py
+++ b/ai_edge_torch/generative/utilities/stable_diffusion_loader.py
@@ -743,7 +743,10 @@ class DiffusionModelLoader(BaseLoader):
 
     attention_config = layers_config.AttentionConfig(
         num_heads=config.transformer_num_attention_heads,
-        head_dim=config.block_out_channels[0] // config.transformer_num_attention_heads,
+        head_dim=(
+            config.block_out_channels[0]
+            // config.transformer_num_attention_heads
+        ),
         num_query_groups=config.transformer_num_attention_heads,
         rotary_percentage=0.0,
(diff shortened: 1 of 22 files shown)


ai_edge_torch/generative/utilities/stable_diffusion_loader.py
```diff
@@ -743,6 +743,10 @@ def load(
 
     attention_config = layers_config.AttentionConfig(
```
yichunk, line 744: attnetion_config should have the same update as the one in diffusion.py.
Though in the model loader, the `head_dim` is never used, so the variable here doesn't matter.
Either add a comment for reminding this is not the actual config used in the model, or update to per layer config here

yichunk (commented):
LGTM for stable diffusion part.

ai_edge_torch/generative/utilities/stable_diffusion_loader.py
haozha111: +1

haozha111 (dismissed):

ai_edge_torch/generative/utilities/stable_diffusion_loader.py
talumbau: Thanks @yichunk. I have just now fixed this with an additional commit. @haozha111 if you can LGTM again I willl merge in these changes.

PR #172, opened 2024-08-27
Fix batch norm when weight/bias is None

### LINKED ISSUE #162: Conversion with BatchNorm "affine" disabled fails
### Description of the bug:

I have model with "conv -> batchnorm -> relu" blocks and it fails to convert if I set "affine" to False in BatchNorm

This should reproduce the issue:
```python
import torch.nn as nn
import os
import torch
import ai_edge_torch

os.environ["PJRT_DEVICE"]="CPU"

class TestModel(nn.Module):

    def __init__(self, in_channels, out_channels):
        super().__init__()
        self.layer = self.layer = nn.Sequential(
            nn.Conv2d(in_channels, out_channels, 3, padding = 1, stride=1, dilation=1, bias = 1),
            nn.BatchNorm2d(out_channels, affine=False)
        )
		
    def forward(self, x):
        return self.layer(x)

model = TestModel(128, 128).eval()
x = torch.randn(1, 128, 128, 128)

edge_model = ai_edge_torch.convert(model, (x,))
edge_model.export("test.tflite")
```

Error:
```
  File ".../site-packages/ai_edge_torch/_convert/fx_passes/optimize_layout_transposes_pass/layout_rewrite.py", line 334, in batch_norm
    return a / b * weight + bias, None, None
  TypeError: unsupported operand type(s) for *: 'FunctionalTensor' and 'NoneType'
```

### Actual vs expected behavior:

_No response_

### Any other information you'd like to share?

_No response_

spacycoder:
Fixes #162

### DIFF
--- a/ai_edge_torch/_convert/fx_passes/optimize_layout_transposes_pass/layout_rewrite.py
+++ b/ai_edge_torch/_convert/fx_passes/optimize_layout_transposes_pass/layout_rewrite.py
@@ -331,7 +331,12 @@ def _aten__native_batch_norm_legit_no_training(node):
   def batch_norm(input, weight, bias, running_mean, running_var, momentum, eps):
     a = input - running_mean
     b = torch.sqrt(running_var + eps)
-    return a / b * weight + bias, None, None
+    out = a / b
+    if weight is not None:
+      out = out * weight
+    if bias is not None:
+      out = out + bias
+    return out, None, None
 
   node.target = batch_norm
 


chunnienc (approved):

chunnienc:
Model coverage tests is not available in for github PRs now. I'm running the tests locally.

Overall the fix LGTM. Thank you very much for debugging and sending us the fix!

PR #687, opened 2025-05-28
[fix] Fix export to path without directory

gudgud96:
`.export("xxx.tflite")` throws an error as it expects the path to contain a directory and creates it (introduced by [this commit](https://github.com/google-ai-edge/ai-edge-torch/commit/c74b0fa3a79469b92fea6c1b45157d7a0d3e8b24)).

Fixed by checking if the path contains a directory, and only creates it if there is.


chunnienc (dismissed):
Thanks for the PR!

Issue #870, opened 2025-12-07
can only convert model using static shapes


stevenctl:
### Description of the bug:

Like several other users found, exporting with dynamic shapes doesn't seem to work. I'm filing a separate issue as the errors I get aren't quite the same.

I've tested this on `0.7.0` as well as on the nightly

```
pip list | grep ai-edge
ai-edge-litert            2.0.3
ai-edge-litert-nightly    2.1.0rc2.dev20251205
ai-edge-quantizer         0.4.0
ai-edge-quantizer-nightly 0.5.0.dev20251207
ai-edge-torch             0.8.0
```

I'm trying to export an ALBERT model (impl here: https://github.com/huggingface/transformers/tree/main/src/transformers/models/albert)

I've tested that dynamic sequence length and batch size works fine both with an exported torch snapshot _and_ on my ONNX export of the same model. 

If I try to export this model with _only_ sequence length dynamic:

```
✗ Export to out/albert-zh-ai-edge-only_seq_len_dynamic/albert-zh.tflite failed: Attempted boolean conversion of traced array with shape bool[].
The error occurred while tracing the function lower_wrapper at /home/landow/gamedev/chinese_app/data/word_seg/autopinyin/venvagain/lib/python3.11/site-packages/ai_edge_torch/odml_torch/jax_bridge/_wrap.py:78 for jit. This concrete value was not available in Python because it depends on the value of the argument args[1].
See https://docs.jax.dev/en/latest/errors.html#jax.errors.TracerBoolConversionError

While executing %slice_1 : [num_users=1] = call_function[target=torch.ops.aten.slice.Tensor](args = (%b_albert_embeddings_token_type_ids, 1, 0, %sym_size_int_83), kwargs = {})
Original traceback:
  File "/home/landow/gamedev/chinese_app/data/word_seg/autopinyin/venvagain/lib/python3.11/site-packages/transformers/models/albert/modeling_albert.py", line 1124, in forward
    outputs = self.albert(
  File "/home/landow/gamedev/chinese_app/data/word_seg/autopinyin/venvagain/lib/python3.11/site-packages/transformers/models/albert/modeling_albert.py", line 684, in forward
    buffered_token_type_ids = self.embeddings.token_type_ids[:, :seq_length]

Use tlparse to see full graph. (https://github.com/pytorch/tlparse?tab=readme-ov-file#tlparse-parse-structured-pt2-logs)
```

or if I add batch size

```

[...]

PR #1023, opened 2026-05-13
Preserve dict keys in flat_arg_names for positional args

### LINKED ISSUE #1022: Dict input keys are flattened to generic args_0/args_1, losing original names
**Describe the bug**
When a `torch.nn.Module` accepts a `dict` as its forward argument, `litert_torch.convert` flattens the dict values into positional tensors and renames them to generic `args_0`, `args_1`, etc. This loses the original dict keys that `torch.export.export` preserves.

**To Reproduce**
[gist](https://colab.research.google.com/gist/pctablet505/3108d4cf47f5e7d846c65e3d52826c9d/litert_torch_input_name_lost.ipynb#scrollTo=JYlGtXTnHo6n)

**Expected behavior**
Signature inputs should retain the original dict keys (`inputs_x`, `inputs_y`) instead of being renamed to `args_0` / `args_1`.

**Environment**
- litert-torch 0.9.0
- ai-edge-litert 1.2.0
- torch 2.6.0
- Python 3.13

**Additional context**
The root cause appears to be in `litert_torch/_convert/signature.py` where `flat_arg_names` generates generic `args_{i}` names from pytree flattening, ignoring the dict keys available in the `ExportedProgram`'s `graph_signature.input_specs`. This makes it hard to map inputs back to their semantic meaning when running inference on the resulting TFLite model.

As a side note, `litert_torch` itself triggers `FutureWarning: treespec.children_specs is deprecated` internally during conversion (see `signature.py:52` and `signature.py:58`), suggesting the pytree traversal code may also need updating to the newer `treespec.children()` API.

pctablet505:
**Problem**
When a `torch.nn.Module` accepts a `dict` as a positional forward argument, `litert_torch.convert` flattens the dict values but renames them to generic `args_0`, `args_1`, etc. This loses the original dict keys that `torch.export.export` preserves.

In addition, the codebase uses the deprecated `treespec.children_specs` API which emits `FutureWarning` on PyTorch 2.6+.

**Fixes in this PR**
1. **Preserve dict keys in `flat_arg_names`** (`signature.py`)
   - Uses `flat_dict_names` for positional args too (with an `args_` prefix), so dict keys are preserved.
   - Simple tensor positional args still produce `args_0`, `args_1` (backward compatible).

2. **Fix deprecated pytree API usage** (`export_utils.py`, `litert_converter.py`)
   - Replaces all `treespec.children_specs` with `treespec.children()`.

3. **Guard `in_i32` against non-integer args** (`export.py`)
   - When dynamic shapes are present, `arange` ops may have symbolic `start`/`end` values. The helper now skips the rewrite gracefully instead of crashing with `TypeError`.

4. **Fail fast on `dynamic_shapes`** (`interface.py`)
   - The JAX-based lowering path does not yet support dynamic dimensions. Rather than crashing with obscure JAX assertion errors, `convert()` now raises a clear `NotImplementedError` pointing to issue #870.

Fixes #1022
Related to #870

### DIFF
--- a/litert_torch/_convert/signature.py
+++ b/litert_torch/_convert/signature.py
@@ -49,16 +49,16 @@ class Signature:
   @property
   def flat_arg_names(self) -> list[str]:
     spec = pytree.tree_flatten(self._normalized_sample_args_kwargs)[1]
-    args_spec, kwargs_spec = spec.children_specs
-    names = []
-    for i in range(args_spec.num_leaves):
-      names.append(f"args_{i}")
+    args_spec, kwargs_spec = spec.children()
+    args_names = backend.export_utils.flat_dict_names(
+        args_spec.children(), args_spec.context
+    )
+    args_names = [f"args_{name}" for name in args_names]
 
     kwargs_names = backend.export_utils.flat_dict_names(
-        kwargs_spec.children_specs, kwargs_spec.context
+        kwargs_spec.children(), kwargs_spec.context
     )
-    names.extend(kwargs_names)
-    return names
+    return args_names + kwargs_names
 
   @property
   def flat_args(self) -> tuple[Any, ...]:
--- a/litert_torch/backend/export_utils.py
+++ b/litert_torch/backend/export_utils.py
@@ -28,7 +28,7 @@ IR_DYNAMIC = -9223372036854775808
 
 
 def flat_dict_names(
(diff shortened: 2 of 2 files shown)


pctablet505:
Closing this PR as stale and conflicting with upstream.

PR #1063, opened 2026-06-19
Fix constant aliasing in inline_consts caused by data_ptr() cache key

### LINKED ISSUE #1061: conv_transpose1d weight constants deduplicated by shape — 2nd op silently reuses 1st's weights
### Summary

Two `F.conv_transpose1d` ops whose weight **constants** share the same **shape** but have
**different values** are deduplicated during conversion: the second op silently reuses the
first op's weights. The trigger is weight-**value** dependent — the inverse-DFT cos/sin basis
pair (a normal inverse-STFT weight pair) gets merged, while random same-shape weights do not.

I hit this porting Kokoro-82M (TTS, ISTFTNet vocoder): it corrupts an in-graph inverse-STFT.
It reproduces on the default delegate **and** the reference kernels
(`BUILTIN_WITHOUT_DEFAULT_DELEGATES`), so it is conversion-level, not a runtime/XNNPACK issue.

### Environment

`torch 2.12.0` · `litert_torch 0.9.1` · `ai_edge_litert 2.1.5` · Python 3.11 · macOS arm64.

### Minimal repro

```python
import os
import numpy as np
import torch
import torch.nn as nn
import torch.nn.functional as F
import litert_torch
from ai_edge_litert.interpreter import Interpreter, OpResolverType


def corr(a, b):
    a, b = np.asarray(a).ravel(), np.asarray(b).ravel()
    n = min(len(a), len(b))
    return float(np.corrcoef(a[:n], b[:n])[0, 1])


def dft_inverse_bases(n_fft=20):
    """Inverse-DFT cos/sin bases (Hann window), each [n_fft//2+1, 1, n_fft]."""
    fb = n_fft // 2 + 1
    win = torch.hann_window(n_fft, periodic=True, dtype=torch.float32).numpy()
    n, k = np.arange(n_fft), np.arange(fb)
    ang = 2 * np.pi * np.outer(n, k) / n_fft
    inv_win = win * (1.0 / n_fft)
    W_cos = torch.from_numpy((np.cos(ang).T * inv_win)).float().unsqueeze(1)
    W_sin = torch.from_numpy((np.sin(ang).T * inv_win)).float().unsqueeze(1)
    return W_cos, W_sin


class TwoConvT(nn.Module):
    """out = conv_transpose1d(a, W1) - conv_transpose1d(b, W2)"""
    def __init__(self, W1, W2, stride):
        super().__init__()
        self.register_buffer("W1", W1)
        self.register_buffer("W2", W2)
        self.stride = stride

    def forward(self, a, b):
        return (F.conv_transpose1d(a, self.W1, stride=self.stride)
                - F.conv_transpose1d(b, self.W2, stride=self.stride)).squeeze()


def convert_and_run(m, args, rt):
    path = "/tmp/convT_dedup.tflite"
    if os.path.exists(path):
        os.remove(path)
    litert_torch.convert(m, tuple(args)).export(path)
    it = Interpreter(model_path=path, experimental_op_resolver_type=rt)
    it.allocate_tensors()
    for d, x in zip(it.get_input_details(), args):
        it.set_tensor(d["index"], x.numpy().astype(d["dtype"]))
    it.invoke()
    return it.get_tensor(it.get_output_details()[0]["index"])


def check(name, W1, W2, stride, in_ch, length):
    torch.manual_seed(0)
    a, b = torch.randn(1, in_ch, length), torch.randn(1, in_ch, length)
    m = TwoConvT(W1, W2, stride).eval()
    with torch.no_grad():
        correct = m(a, b).numpy()
        wrong = (F.conv_transpose1d(a, W1, stride=stride)
                 - F.conv_transpose1d(b, W1, stride=stride)).squeeze().numpy()  # 2nd reuses W1
    print(f"\n== {name}  weights={tuple(W1.shape)} stride={stride} ==")
    for rt in (OpResolverType.AUTO, OpResolverType.BUILTIN_WITHOUT_DEFAULT_DELEGATES):
        tfl = convert_and_run(m, (a, b), rt)
        c_ok, c_bug = corr(tfl, correct), corr(tfl, wrong)
        verdict = "BUG (2nd conv reuses 1st weights)" if c_bug > 0.999 and c_ok < 0.99 else "ok"
        print(f"   {str(rt).split('.')[-1]:34s} corr_vs_correct={c_ok:.4f}  corr_vs_reused-W1={c_bug:.4f}  -> {verdict}")


W_cos, W_sin = dft_inverse_bases(20)
check("CASE A (DFT cos/sin bases)", W_cos, W_sin, stride=5, in_ch=11, length=200)
torch.manual_seed(1)
check("CASE B control (random, same shape)", torch.randn(11, 1, 20), torch.randn(11, 1, 20),
      stride=5, in_ch=11, length=200)
```

### Observed

```
== CASE A (DFT cos/sin bases)  weights=(11, 1, 20) stride=5 ==
   AUTO                               corr_vs_correct=0.5587  corr_vs_reused-W1=1.0000  -> BUG (2nd conv reuses 1st weights)
   BUILTIN_WITHOUT_DEFAULT_DELEGATES  corr_vs_correct=0.5587  corr_vs_reused-W1=1.0000  -> BUG (2nd conv reuses 1st weights)
== CASE B control (random, same shape)  weights=(11, 1, 20) stride=5 ==
   AUTO                               corr_vs_correct=1.0000  corr_vs_reused-W1=0.5077  -> ok
   BUILTIN_WITHOUT_DEFAULT_DELEGATES  corr_vs_correct=1.0000  corr_vs_reused-W1=0.5077  -> ok
```

The converted CASE-A output matches "both convs use W_cos" at corr **1.0000** (vs 0.56 to the
correct `cos − sin`). CASE B (random, identical shapes) converts correctly — so a constant-dedup
pass is incorrectly collapsing the two distinct (DFT-structured) weight constants.

### Impact

Corrupts any in-graph inverse-STFT / vocoder / dual-branch conv built from a shared-shape
weight pair (very common in TTS/audio). For Kokoro-82M I worked around it by moving the final
iSTFT host-side (numerically exact, ~free), but the in-graph path is silently wrong.

### Expected

Both cases convert to `corr_vs_correct ~1.0`; weight constants must not be merged by value/shape.

john-rocky:
## Summary

`inline_consts._tensor_fingerprint` keyed the constant cache on
`tensor.untyped_storage().data_ptr()`, computed on the temporary copy produced
by `x.contiguous().detach().cpu()`. That temporary is freed when the lowering
returns, and the allocator reuses its address for the next constant — a false
cache hit that silently merges two **distinct** constants. This corrupts any
in-graph inverse-STFT / vocoder: the cos/sin inverse-DFT `conv_transpose1d`
weights collide and the second op reuses the first's MLIR attribute.

Root-cause analysis with instrumented proof is in
#1061 (https://github.com/google-ai-edge/litert-torch/issues/1061#issuecomment-4749470635).

## Fix

Key the cache on tensor **content** (blake2b-128 over the raw bytes,
reinterpreted via `view(torch.uint8)` so it is dtype-agnostic, incl. bfloat16)
instead of a storage address. This also makes the intended cross-signature
deduplication more correct: genuinely identical constants now share even across
distinct tensor objects.

## Test

Adds `litert_torch/backend/test/test_inline_consts.py`:

- `test_recycled_storage_with_different_content_differs` — same storage /
  `data_ptr()`, different content must yield different fingerprints (this fails
  on the old key).
- `test_equal_content_distinct_storage_matches` — distinct storage, identical
  content still deduplicates.
- `test_distinct_conv_transpose_weights_not_aliased` — two distinct
  `conv_transpose1d` weights (cos/sin inverse-DFT bases) lower to two distinct
  constants.

Verified locally: the #1061 repro now reports `corr_vs_correct=1.0` for the
cos/sin case; the random-weight control is unaffected; the new tests and the
existing backend tests pass.

Fixes #1061

### DIFF
--- a/litert_torch/backend/inline_consts.py
+++ b/litert_torch/backend/inline_consts.py
@@ -15,6 +15,7 @@
 """The pre-lower pass to make exported program's constant inputs into MLIR ElementsAttrs."""
 
 import dataclasses
+import hashlib
 import logging
 import math
 from litert_torch import _config
@@ -30,22 +31,35 @@ config = _config.config
 
 @dataclasses.dataclass(frozen=True)
 class _ConstantFingerprint:
-  """Unique fingerprint of a torch.Tensor constant for caching."""
+  """Unique content-based fingerprint of a torch.Tensor constant for caching.
+
+  The fingerprint must NOT be keyed on a storage address such as
+  ``tensor.untyped_storage().data_ptr()``: the storage backing one constant
+  can be freed and the same address later reused by a different, unrelated
+  constant, producing a false cache hit that silently aliases two distinct
+  constants. See https://github.com/google-ai-edge/litert-torch/issues/1061.
+  """
 
   device: str
   shape: tuple[int, ...]
-  stride: tuple[int, ...]
-  data_ptr: int
-  storage_offset: int
+  dtype: str
+  digest: str
 
   @classmethod
   def from_tensor(cls, tensor: torch.Tensor) -> '_ConstantFingerprint':
+    flat = tensor.detach().cpu().contiguous().reshape(-1)
+    if flat.numel():
+      # Reinterpret as raw bytes so the hash is dtype-agnostic (works for
+      # dtypes numpy cannot represent directly, e.g. bfloat16).
+      flat = flat.view(torch.uint8)
+    digest = hashlib.blake2b(
+        memoryview(flat.numpy()).cast('B'), digest_size=16
+    ).hexdigest()
     return cls(
         device=str(tensor.device),
-        shape=tensor.shape,
-        stride=tensor.stride(),
-        data_ptr=tensor.untyped_storage().data_ptr(),
-        storage_offset=tensor.storage_offset(),
+        shape=tuple(tensor.shape),
+        dtype=str(tensor.dtype),
+        digest=digest,
     )
 
 
--- a/litert_torch/backend/test/test_inline_consts.py
+++ b/litert_torch/backend/test/test_inline_consts.py
@@ -1,0 +1,127 @@
+# Copyright 2026 The LiteRT Torch Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Tests for the inline-constants pre-lower pass.
+
+Regression coverage for https://github.com/google-ai-edge/litert-torch/issues/1061:
+two distinct constants must never be merged by the constant cache, even when one
+constant's storage is freed and its address reused by another.
+"""
+
+import numpy as np
+import torch
+import torch.nn.functional as F
+
+from litert_torch import backend
+from litert_torch.backend import inline_consts
+from litert_converter.mlir import ir
+
+from absl.testing import absltest as googletest
+
+
+class TestTensorFingerprint(googletest.TestCase):
+  """The constant-cache key must depend on content, not on a storage address."""
+
+  def test_recycled_storage_with_different_content_differs(self):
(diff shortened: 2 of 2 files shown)


outtanames:
@john-rocky can you resolve the conflicts on this

john-rocky:
Thanks for the ping — resolved by rebasing onto main. The fix now lives in the `_ConstantFingerprint` dataclass from main's refactor: the cache key is a blake2b content digest instead of the storage address. The #1061 regression tests pass on the rebased branch.

Issue #1060, opened 2026-06-19
Conversion fails: torch.uint8 unsupported in MLIR lowering (blocks Qwen3.5 / GatedDeltaNet)


john-rocky:
### Summary

The MLIR/StableHLO backend has no mapping for `torch.uint8`, so any `uint8` tensor that
survives to lowering aborts conversion with `RuntimeError: torch.uint8`. I hit this trying
to convert **Qwen3.5** (the new GatedDeltaNet hybrid) through litert_torch's own `qwen3_5`
export builder — the `chunk_gated_delta_rule` path emits a `uint8` `arange` for chunk
indexing and conversion dies.

### Environment

`torch 2.12.0` · `litert_torch 0.9.1` · `transformers 5.12.1` · `ai_edge_litert 2.1.5` ·
Python 3.11 · macOS arm64 (M-series).

### Minimal repro

The same graph converts with `int32` and fails with `uint8`, so it's the dtype in the
backend, not the `arange` op or the model:

```python
import torch
import torch.nn as nn
import litert_torch


class ArangeAdd(nn.Module):
    """y = x + arange(0, N, dtype=DT).float() — the arange survives to lowering
    because it is consumed in arithmetic with the (dynamic) input."""
    def __init__(self, dtype):
        super().__init__()
        self.dtype = dtype

    def forward(self, x):
        idx = torch.arange(0, x.shape[-1], dtype=self.dtype)   # the carrier op
        return x + idx.to(torch.float32)


def try_convert(dtype):
    try:
        litert_torch.convert(ArangeAdd(dtype).eval(), (torch.randn(1, 64),))
        return None
    except Exception as e:
        return f"{type(e).__name__}: {str(e).splitlines()[0]}"


for label, dt in [("uint8", torch.uint8), ("int32 (control)", torch.int32)]:
    err = try_convert(dt)
    print(f"{label:18s} -> {'FAILED: ' + err if err else 'converted OK'}")
```

Output:

```
uint8              -> FAILED: RuntimeError: torch.uint8
int32 (control)    -> converted OK
```

### Where it bites in the real model

Converting `Qwen/Qwen3.5-0.8B` (also 2B / 4B / 9B — all GatedDeltaNet hybrids) via the
dedicated `qwen3_5` builder fails at the same op:

```
RuntimeError: torch.uint8
While executing %arange_71 = aten.arange.start_step(0, 64,
    kwargs={dtype: torch.uint8, layout: torch.strided, device: cpu})

[...]

PR #1064, opened 2026-06-19
Map ui8 result type back to torch.uint8 in ir_element_type_to_torch_dtype

john-rocky:
## Summary

`deef131e` ("Add support for torch.uint8 dtype.") added the forward
`torch.uint8 -> IntegerType.get_unsigned(8)` mapping, but the reverse mapping in
`ir_element_type_to_torch_dtype` was left broken: the `torch.uint8` branch is
nested under `if ty.is_signless:`, and an unsigned integer type has
`is_signless == False`, so the branch is unreachable. A `ui8` type therefore
falls through to `RuntimeError: Unsupported ir element type: ui8` — reached e.g.
via `jax_bridge/utils.py`, which maps bridged-op result types back to torch
dtypes. Background and a minimal repro are in #1060.

## Fix

Lift the unsigned check to be a sibling of the `is_signless` check, so a `ui8`
type maps back to `torch.uint8`. This completes the reverse direction of the
`torch.uint8` support added in `deef131e`.

## Test

Adds `litert_torch/backend/test/test_export_utils.py`:
- a parameterized `torch_dtype <-> ir element type` round-trip across all
  supported dtypes (including uint8);
- an explicit `ui8 -> torch.uint8` regression.

Verified: on `main` the new tests fail with `Unsupported ir element type: ui8`;
with this change they pass.

Refs #1060

### DIFF
--- a/litert_torch/backend/export_utils.py
+++ b/litert_torch/backend/export_utils.py
@@ -212,10 +212,10 @@ def ir_element_type_to_torch_dtype(ty):
   if isinstance(ty, ir.F16Type):
     return torch.half
   if isinstance(ty, ir.IntegerType):
+    if ty.is_unsigned:
+      if ty.width == 8:
+        return torch.uint8
     if ty.is_signless:
-      if ty.is_unsigned:
-        if ty.width == 8:
-          return torch.uint8
       if ty.width == 64:
         return torch.long
       if ty.width == 32:
--- a/litert_torch/backend/test/test_export_utils.py
+++ b/litert_torch/backend/test/test_export_utils.py
@@ -1,0 +1,61 @@
+# Copyright 2024 The LiteRT Torch Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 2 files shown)


john-rocky:
Now that this is up for review — a self-contained summary; the longer history is in #1060.

`deef131e` ("Add support for torch.uint8 dtype.") added the uint8 mapping in both directions, but the reverse-direction lines in `ir_element_type_to_torch_dtype` (`litert_torch/backend/export_utils.py`) landed nested under `if ty.is_signless:`. An MLIR `IntegerType` is signless, signed, or unsigned — never two at once (`ui8` reports `is_unsigned=True, is_signless=False`) — so those lines are unreachable. A `ui8` result type therefore falls through to `RuntimeError: Unsupported ir element type: ui8`. This PR moves the same 3 lines up one level, to a sibling of the `is_signless` block; the second file is a round-trip test over 8 dtypes plus a uint8 regression case.

To see it in one run: `python litert_torch/backend/test/test_export_utils.py`. On current `main` (`d49f1ba`) plus this PR's test file, the two uint8 cases raise `Unsupported ir element type: ui8` and the other 7 pass; on this branch, 9/9 pass (both re-run today).

Practical impact is in #1060: it blocks Qwen3.5 / GatedDeltaNet conversion (uint8-typed result tensors).

Happy to rebase or reshape anything that doesn't fit.

PR #1079, opened 2026-06-23
Keep attention QKV-unpack view chain within rank 4 for the GPU delegate

john-rocky:
## Summary

A model that uses `torch.nn.MultiheadAttention` for self-attention — which also
covers PyTorch's built-in `nn.TransformerEncoder` / `nn.TransformerEncoderLayer`,
and e.g. CLIP / OpenCLIP — converts to a `.tflite` whose desktop op-set looks fine,
but on the **ML Drift GPU delegate (`LITERT_CL`)** the whole `CompiledModel` fails
to compile.

Root cause: torch's `_in_projection_packed` unpacks the packed QKV projection
with an `unsqueeze -> permute -> squeeze` view chain that **transiently
materializes rank-5 tensors**, even though the inserted dimension is size 1 and
is removed again immediately. ML Drift caps tensors at rank 4, so these
transient rank-5 `RESHAPE`/`TRANSPOSE` intermediates push every attention block
off the GPU and the partitioner cannot place the graph. There is no CPU fallback
in `CompiledModel(Accelerator.GPU)`, so the model never loads:

```
E tflite : Following operations are not supported by GPU delegate:
E tflite : RESHAPE: Tensor "...MultiheadAttention_attn;3" has bad input dims size: 5.
E tflite : RESHAPE: Tensor "...MultiheadAttention_attn;4" has bad input dims size: 5.
I tflite : Replacing 65 out of 751 node(s) with delegate (LITERT_CL) node ...
E litert : Failed to create compiled model: Failed to compile model
```

## The fix

`unsqueeze`, `permute` and `squeeze` are all pure metadata views, so the chain
`squeeze(permute(unsqueeze(x, d), perm), sq_dims)` is **exactly equal to a single
`permute` of `x`** (optionally followed by a `squeeze` of any other unit dims),
which never exceeds the rank of `x`.

This PR adds `ReduceViewRankPass`, which performs that algebraic folding. It is a
numerical no-op, and it **only fires when the `unsqueeze` actually inflates a
tensor past rank 4**, so graphs that are already within the rank-4 cap are left
untouched.

After this pass, stock `nn.MultiheadAttention` converts with **max tensor rank 4
(was 5, two rank-5 tensors)** on the generic `litert_torch.convert` path, with no
model-side change.

## Verification (reproduced on `main` HEAD `771c7bb`)

A tiny `nn.MultiheadAttention` model, converted via `litert_torch.convert` and
inspected with `ai_edge_litert.Interpreter`:

| | max tensor rank | #tensors rank>4 |
|---|---|---|
| before | 5 | 2 |
| after  | 4 | 0 |

- **Numerically exact:** the converted `.tflite` matches eager PyTorch to
  `max|diff| ≈ 1e-7` (float32 conversion noise) across `batch_first` True/False,
  multiple head/embed sizes, batch>1, and stacked attention blocks.
- **No regression:** a non-attention conv net is byte-for-byte unaffected (the
  pass does not fire); `litert_torch/_convert/test/test_convert.py` passes (26
  tests). A new `test_reduce_view_rank_pass.py` covers the fold, the numeric
  equivalence, `batch_first=False`, the multi-unit-dim squeeze branch, and the
  no-op case.

### On the full CLIP ViT-B/32 encoder

Converting the real **open_clip ViT-B/32** visual encoder through
`litert_torch.convert` (same path, same `main` HEAD):

| | max tensor rank | #tensors rank>4 |
|---|---|---|
| before | 5 | **24** (the two packed-QKV reshapes per `ResidualAttentionBlock`, ×12) |
| after  | 4 | **0** |

The 24 rank-5 tensors before the fix are exactly the ones the GPU delegate
rejects — they carry names like
`...ResidualAttentionBlock_0/...MultiheadAttention_attn;1` / `;2`, shapes
`(3,50,1,1,768)` / `(1,50,1,3,768)`. The pass removes all of them and the encoder
output stays faithful (`max|diff| ≈ 2.6e-6` tflite vs eager).

## On-device measurement (Pixel 8a, LITERT_CL / ML Drift)

I converted the pretrained (`openai`) open_clip ViT-B/32 image encoder twice — once
on pristine `main` HEAD and once with this PR, both with **stock
`nn.MultiheadAttention` and no model-side attention rewrite** — and loaded each on a
Pixel 8a via the LiteRT `CompiledModel(Accelerator.GPU)` path. The two `.tflite`s are
identical except for the litert-torch version:

| litert-torch | GPU delegate (`LITERT_CL`) | `CompiledModel(GPU)` |
|---|---|---|
| `main` HEAD (before) | `Replacing 65 out of 751 node(s) … 2 partitions`, then `RESHAPE … has bad input dims size: 5` on `ResidualAttentionBlock_*/…MultiheadAttention` | **Failed to compile model** → CPU fallback |
| this PR (after) | `Replacing 739 out of 739 node(s) … 1 partition` | **GPU ready** (whole encoder on the GPU) |

So the rank-4 attention lowering **alone** flips the encoder from "can't compile on
the GPU delegate" to "entire graph delegated", with no NHWC/layout change. (The node
totals differ — 751 vs 739 — because the pass removes the redundant
unsqueeze/permute/squeeze ops.) The fixed encoder also still classifies correctly
end-to-end (zero-shot top-1 matches stock CLIP), so it's GPU-clean *and* faithful.

## Surface question

I went with a dedicated FX pass because the rank-5 comes from a **multi-op view
chain** (torch's packed-QKV unpack), not a single decomposable aten op, so the
per-op composite/decomp-table mechanisms don't apply cleanly here. I placed it
right after `CanonicalizePass` in `_run_convert_passes`.

If you'd prefer a different surface (e.g. folding this into `CanonicalizePass`,
or a narrower/broader matcher), I'm happy to reshape it — just let me know which
you'd like.

**Scope note (what this does and does not cover):** this folds the rank-5 chain
that torch's `nn.MultiheadAttention` packed-QKV projection emits
(`unsqueeze → permute → squeeze`). Verified covered: `nn.MultiheadAttention`
self-attention and PyTorch's `nn.TransformerEncoder`/`nn.TransformerEncoderLayer`.
Already fine (no rank-5, untouched): attentions that project q/k/v separately and
split per head with `view(B, N, H, d).transpose(...)` (the typical HF-transformers
shape). **Not yet covered:** attentions that build packed QKV with a manual reshape
— `qkv.reshape(B, N, 3, H, d).permute(...).unbind(0)`, as used by timm ViTs,
DINO/DINOv2, many DETR-style models and Depth Anything 3 — which hit the same
rank-5 wall through a different `view → permute → slice` chain that this matcher
doesn't fold yet. Extending it to that chain (keeping the whole ViT family
GPU-clean) is a natural, high-value follow-up — happy to fold it into this PR or do
a follow-up, whichever you prefer.

### DIFF
--- a/litert_torch/_convert/fx_passes/test/test_reduce_view_rank_pass.py
+++ b/litert_torch/_convert/fx_passes/test/test_reduce_view_rank_pass.py
@@ -126,6 +126,26 @@ class TestReduceViewRankPass(googletest.TestCase):
     self.assertLessEqual(max_rank, 4)
     torch.testing.assert_close(result.graph_module(x), f(x))
 
+  def test_permute_with_second_user_is_not_folded(self):
+    # The permute output forks to a second consumer besides the squeeze, so
+    # the rank-5 chain must stay alive for that branch; the pass backs off.
+    def f(x):  # x: (2, 1, 3, 4)
+      y = torch.unsqueeze(x, 0)  # (1, 2, 1, 3, 4)
+      y = torch.permute(y, [1, 0, 2, 3, 4])  # (2, 1, 1, 3, 4)
+      z = torch.squeeze(y, 1)  # (2, 1, 3, 4)
+      return z, torch.sum(y)  # second user of the permute
+
+    x = torch.randn(2, 1, 3, 4)
+    graph_module = make_fx(f, tracing_mode="fake")(x)
+    targets_before = [n.target for n in graph_module.graph.nodes]
+
+    result = fx_passes.ReduceViewRankPass()(graph_module)
+
+    self.assertFalse(result.modified)
+    targets_after = [n.target for n in result.graph_module.graph.nodes]
+    self.assertEqual(targets_before, targets_after)
+    torch.testing.assert_close(result.graph_module(x), f(x))
+
   def test_provenance_meta_copied_to_new_nodes(self):
     # The folded permute/squeeze nodes must carry the debug provenance of the
     # squeeze node they replace.


john-rocky:
Added a from-scratch on-device A/B (Pixel 8a, `LITERT_CL` / ML Drift GPU) to the description.

I converted the pretrained (`openai`) open_clip ViT-B/32 image encoder twice — **stock `nn.MultiheadAttention`, no model-side rewrite** — on pristine `main` HEAD vs this branch, so the only difference is the litert-torch version, then loaded each via `CompiledModel(Accelerator.GPU)`:

- **before (main):** `Replacing 65 out of 751 node(s) … 2 partitions`, then `RESHAPE … has bad input dims size: 5` on `ResidualAttentionBlock_*/…MultiheadAttention` → **Failed to compile model** → CPU fallback.
- **after (this PR):** `Replacing 739 out of 739 node(s) … 1 partition` → **GPU ready**, the whole encoder runs on the GPU delegate (and still classifies correctly end-to-end).

So the rank-4 attention lowering alone flips the encoder from "can't compile on GPU" to "fully delegated", with no NHWC/layout change. The change is a dedicated FX pass placed after `CanonicalizePass` — happy to reshape the surface (fold into `CanonicalizePass`, narrower/broader matcher, etc.) to whatever you prefer.

litert_torch/_convert/fx_passes/reduce_view_rank_pass.py
```diff
@@ -0,0 +1,154 @@
...
+          aten.squeeze.dims, (new_permute, remaining_squeeze)
+      )
+    else:
+      result = new_permute
+
+  # The folded result is element-for-element identical to the squeeze output,
+  # so reuse its metadata.
+  result.meta["val"] = squeeze_node.meta["val"]
+  if "tensor_meta" in squeeze_node.meta:
+    result.meta["tensor_meta"] = squeeze_node.meta["tensor_meta"]
```
outtanames, line 147: Should we also copy the provenance keys from squeeze_node onto the new nodes for debugability?

john-rocky: Thanks for the suggestion — agreed, keeping the debug provenance on the folded nodes is well worth it. Done in 98c8915, pushed to this PR: the new nodes now carry the replaced squeeze node's provenance meta, with a regression test. Nothing further is needed from your side unless you'd like the approach changed.

Details, if useful:
- The pass now copies `stack_trace`, `nn_module_stack`, `source_fn_stack`, and `from_node` from the replaced squeeze node onto both created nodes (the folded `permute`, and the extra `squeeze` when other unit dims remain).
- The key set mirrors the one `OptimizeLayoutTransposesPass.get_source_meta` already uses, so the debuginfo build maps the folded nodes back to the original model code the same way as elsewhere.
- Added `test_provenance_meta_copied_to_new_nodes`: it sets sentinel provenance on the squeeze node and asserts both folded nodes carry all four keys after the pass.
- Happy to factor the key list into a shared `fx_infra` helper instead if that fits the codebase better — your call on convention.

litert_torch/_convert/fx_passes/test/test_reduce_view_rank_pass.py
```diff
@@ -0,0 +1,187 @@
+# Copyright 2025 The LiteRT Torch Authors.
```
outtanames: [nit] can we update this to 2026?

outtanames (changes requested):

litert_torch/_convert/fx_passes/test/test_reduce_view_rank_pass.py
john-rocky: Done in 75bb9e0 — updated to 2026 on both new files (the pass and the test). Thanks for rebasing the branch onto main.

outtanames:
Confirming that I am able to reproduce reshape failure: `RESHAPE: Tensor dimensions must be less than 5. __main__.CLIPAttention/torch.nn.modules.activation.MultiheadAttention_attn`.

outtanames (dismissed):
LGTM

litert_torch/_convert/fx_passes/reduce_view_rank_pass.py
```diff
@@ -0,0 +1,170 @@
...
+    return False
+
+  permute_node = next(iter(unsqueeze_node.users))
+  if (
+      permute_node.target != aten.permute.default
+      or len(permute_node.users) != 1
+  ):
+    return False
+
+  squeeze_node = next(iter(permute_node.users))
```
kimishpatel, line 90: dont you need to check if permute_node also has only one user? If not you need to preserve the other branch in the fork?

john-rocky: Thanks for the careful read — this fork case is exactly the right thing to poke at.

That check is in the pass: the condition right above the line you flagged bails out when the permute has more than one user, so a forked permute is never folded ([reduce_view_rank_pass.py#L83-L88](https://github.com/john-rocky/litert-torch/blob/5aaa679b105ecf51cb28decef0a45b336b5b6818/litert_torch/_convert/fx_passes/reduce_view_rank_pass.py#L83-L88)):

```python
permute_node = next(iter(unsqueeze_node.users))
if (
    permute_node.target != aten.permute.default
    or len(permute_node.users) != 1
):
  return False
```

The unsqueeze gets the same single-user check just before it. On preserving the other branch instead: with a second consumer, the rank-5 unsqueeze/permute must stay alive for that branch anyway, so folding just the squeeze path would leave the rank-5 intermediates in the graph and gain nothing — that's why the pass backs off entirely on any fork. The tail squeeze is the one node that needs no user check, since all of its users are rewired to an element-for-element identical value.

The tests didn't cover this case, so I've added `test_permute_with_second_user_is_not_folded` to the PR: it builds exactly that fork (a squeeze plus a second consumer of the permute output) and asserts the pass reports no modification and leaves the graph node-for-node unchanged. I checked that removing the users check makes it fail, so the guard is now pinned by the test.

PR #1087, opened 2026-06-30
Add NoPE (per-layer RoPE disable) support + SmolLM3-3B example

john-rocky:
## Summary

NoPE-interleave models — where rotary positional embedding (RoPE) is applied on
most layers but **switched off on a subset** — cannot currently be expressed.
`DecoderOnlyModel.forward` computes a single RoPE from `block_config(0)` and
passes the same tuple to every block, and the only knob, `rotary_percentage`, is
read from block 0 only. So a model like **SmolLM3**, which disables RoPE on every
4th layer (NoPE), has no representation. This also covers the NoPE-in-global
layers used by OLMo-3 / Arcee-class models.

## Fix

Add `AttentionConfig.enable_rope: bool = True`. When `False`, `CausalSelfAttention`
(and the cross-attention path) skip `apply_rope_inline` for that layer. RoPE
parameters are still taken from a RoPE-enabled layer; the flag only gates whether
RoPE is **applied**, so the global RoPE tuple is unchanged and the default keeps
every existing model bit-identical. It follows the existing per-block config
pattern (`attn_type` / `query_norm_config`), and `block_configs` already accepts a
per-layer `Sequence`.

A NoPE layer strictly *removes* the RoPE `MUL`/reshape ops — no new op is emitted,
so the existing GPU-delegate path is unaffected.

## Example

`examples/smollm/`: adds `SmolLM3` + `get_model_config_v3()` and
`convert_v3_to_tflite.py`. SmolLM3-3B = 36 layers, GQA (16 heads / 4 KV groups),
GATED-SiLU MLP (11008), RMSNorm, `rotary_base=5e6`, vocab 128256, tied embeddings,
**RoPE disabled on layers 3, 7, …, 35** (`enable_rope=False`).

## Test

`layers/attention_nope_test.py`: asserts that a layer with `enable_rope=False`
produces output **identical** to supplying no RoPE, and that with RoPE enabled the
output differs. Passes locally.


litert_torch/generative/examples/smollm/smollm.py
```diff
@@ -119,3 +119,78 @@ def build_model_v2(
...
+      vocab_size=128256,
+      num_layers=num_layers,
+      max_seq_len=65536,
+      embedding_dim=2048,
+      block_configs=[block_config(i) for i in range(num_layers)],
+      final_norm_config=norm_config,
+  )
+
+
+def get_fake_model_config_v3() -> cfg.ModelConfig:
```
outtanames, line 173: Is this used anywhere?

litert_torch/generative/layers/attention_nope_test.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2024 The LiteRT Torch Authors.
```
outtanames: Can we roll this into `attention_test.py` to follow the same pattern?

litert_torch/generative/examples/smollm/smollm.py
```diff
@@ -119,3 +119,78 @@ def build_model_v2(
...
+  config = get_model_config_v3()
+  config.vocab_size = 128
+  # Keep a full NoPE period (layers 0-3; layer 3 is a NoPE layer).
+  config.num_layers = 4
+  config.block_configs = config.block_configs[:4]
+  for i in range(config.num_layers):
+    config.block_config(i).ff_config.intermediate_size = 64
```
outtanames: Is this config value actually per layer?

outtanames (commented):
Do we also need to update `litert_torch/generative/examples/smollm/verify.py` for v3?

litert_torch/generative/examples/smollm/smollm.py
john-rocky: It wasn't — thanks for catching that. Rather than delete it, I wired it into `test_model_conversion_large.py` as `test_smollm3` (same pattern as `test_smollm`/`test_smollm2`) in d76ef9b, so the NoPE-interleaved path now has conversion coverage: the fake config keeps a full 4-layer NoPE period (layer 3 is a NoPE layer). Happy to drop the function instead if you'd rather not grow the large-test suite — your call.

litert_torch/generative/layers/attention_nope_test.py
john-rocky: Done in d76ef9b — rolled into `attention_test.py` as `test_causal_self_attention_nope`, following the existing pattern, and deleted `attention_nope_test.py`.

litert_torch/generative/examples/smollm/smollm.py
john-rocky: No — `intermediate_size` is uniform (11008) across all SmolLM3 layers; only `enable_rope` varies per layer. What made this read as per-layer is a real flaw you spotted: all 36 blocks were sharing a single `FeedForwardConfig` instance, so the loop was rewriting the same object. Fixed in d76ef9b — each block now builds its own `ff_config` (same as `openelm.py`), so per-block edits in the fake config actually stay per-block.

john-rocky:
Thanks for the review! Added in d76ef9b: `v3` is now a `--model_version` option in `verify.py` (checkpoint `HuggingFaceTB/SmolLM3-3B`), and `verify_util._BUILDER` maps it to `build_model_v3`.

john-rocky:
Merged main to bring the branch up to date — no conflicts. Ready for another look whenever it suits.

Issue #1177, opened 2026-08-20
npu_export Stage 2 calibration fails on main: KeyError: 'valid_mask' (cache_update signatures expect an input the sampling executor does not pass)


john-rocky:
Thanks for opening up the end-to-end NPU export pipeline — with the missing `__init__.py` added in b66af07 (Aug 19), the `npu_export` package now imports cleanly from a plain `pip install --pre -e .`, and all four stage dependencies resolve from PyPI. I ran the `litert_lm_npu_export.ipynb` flow as an outside user: Stage 1 (export) and Stage 3 (quantize) work, but Stage 2 (calibrate) fails on any bundle exported from current main, on the first prefill:

```
  File ".../experimental/calib/sampling_executor.py", line 873, in try_get_quantized_input
    input_data = signature_input[k]
                 ~~~~~~~~~~~~~~~^^^
KeyError: 'valid_mask'
```

**Cause — two commits crossed.** bac4c9e (Aug 4, "Add reference ring buffer SWA implementation") added `valid_mask` to `CacheUpdate.forward` and to its sample inputs in `core/split_cache/exportable_module.py`, so every split-cache export now emits `prefill_cache_update_*` / `decode_cache_update` signatures with a `valid_mask` input — also for models without a sliding window. I unpacked the bundle exported at HEAD for Qwen/Qwen3-0.6B and both cache-update signatures do list `valid_mask`. The calibration executor feeds `valid_mask` to the mask runners (guarded by `get_input_details()`, from daad5d3), but not to the two cache-update calls in `prefill_chunk` and `decode_step` — so the lookup of the runner's expected input fails. The NPU pipeline landed Aug 3 (1a46c22), one day before bac4c9e, and the package was not importable before the Aug 19 init fix, which is presumably why this had not surfaced.

**Repro** (macOS arm64, Python 3.12, fresh venv, `pip install --pre -e .` at b66af07; also reproducible from the notebook):

```python
from litert_torch.generative.export_hf.experimental.npu_export.config_manager import build_pipeline_config
from litert_torch.generative.export_hf.experimental.npu_export import stages

cfg = build_pipeline_config(
    model_id="Qwen/Qwen3-0.6B", target_vendor="qualcomm", target_soc="SM8750",
    prefill_lengths=[128], cache_length=1024,
)
stages.npu_export(cfg=cfg, output_dir=EXPORT_DIR)          # OK
stages.npu_calibrate(                                       # KeyError: 'valid_mask'
[...]

PR #1178, opened 2026-08-20
Pass valid_mask to cache_update signatures in calibration executor

john-rocky:
This is the fix proposed in #1177: Stage 2 calibration fails with `KeyError: 'valid_mask'` on any bundle exported from current main, because split-cache exports have emitted `prefill_cache_update_*` / `decode_cache_update` signatures with a `valid_mask` input since bac4c9e, while the sampling executor only fed `valid_mask` to the mask runners.

Both cache-update call sites (`prefill_chunk`, `decode_step`) now pass `valid_mask`, guarded on `get_input_details()` exactly like the existing mask-runner handling — bundles whose cache-update signatures don't have the input keep working unchanged.

**Rebased onto main at `8379afb`, and re-verified as a two-run A/B so the difference is reproducible from a checkout.**

Both runs consume one Stage 1 bundle (Qwen/Qwen3-0.6B, `prefill_lengths=[128]`, `cache_length=1024`, `dynamic_wi8_afp32`) and the same 3-prompt calibration set at 8 decode steps. The only difference is `PYTHONPATH` — a pristine `8379afb` tree, or this branch. Stage 1 is untouched by this PR: the diff changes two methods of the calibration executor, which the export stage never runs, so both runs see the same bundle.

|  | main `8379afb` | this branch |
|---|---|---|
| Stage 2 calibrate | `KeyError: 'valid_mask'` after 2 s, at `sampling_executor.py:525` in `prefill_chunk` | OK in 6 s, 3/3 examples |
| calibration output | 0 files | 4 files |
| Stage 3 quantize | not reached | OK in 7 s, SRQ bundle written (771,346,785 B) |

```
git archive 8379afb | tar -x -C /tmp/main_tree
git archive HEAD    | tar -x -C /tmp/branch_tree
# stage2.py: the npu_calibrate call from #1177, against a bundle exported once
PYTHONPATH=/tmp/main_tree   python stage2.py   # KeyError: 'valid_mask'
PYTHONPATH=/tmp/branch_tree python stage2.py   # calibrate and quantize both complete
```

Environment: macOS arm64, Python 3.12.13, ai-edge-litert-nightly 2.3.0.dev20260819, ai-edge-quantizer-nightly 0.9.0.dev20260820, litert-converter 0.4.0.dev20260819. The full traceback and the signature dump are in #1177.

Happy to reshape this if you'd rather fix it elsewhere (e.g. on the export side).

### DIFF
--- a/litert_torch/generative/export_hf/experimental/calib/sampling_executor.py
+++ b/litert_torch/generative/export_hf/experimental/calib/sampling_executor.py
@@ -522,12 +522,18 @@ class Executor:
         self.prefill_runners[input_size],
     )
 
+    prefill_cache_update_inputs = {
+        **kv_slice,
+        **decode_state.kv_cache,
+        'input_pos': positions,
+    }
+    if (
+        'valid_mask'
+        in self.prefill_cache_update_runners[input_size].get_input_details()
+    ):
+      prefill_cache_update_inputs['valid_mask'] = valid_mask
     new_kv_cache = try_run_signature_with_quant_dequant(
-        {
-            **kv_slice,
-            **decode_state.kv_cache,
-            'input_pos': positions,
-        },
+        prefill_cache_update_inputs,
         self.prefill_cache_update_runners[input_size],
     )
 
@@ -612,12 +618,15 @@ class Executor:
 
     logits = kv_slice.pop('logits')
 
+    decode_cache_update_inputs = {
+        **kv_slice,
+        **decode_state.kv_cache,
+        'input_pos': positions,
+    }
+    if 'valid_mask' in self.decode_cache_update_runner.get_input_details():
+      decode_cache_update_inputs['valid_mask'] = np.ones((1, 1), dtype=np.bool)
     new_kv_cache = try_run_signature_with_quant_dequant(
-        {
-            **kv_slice,
-            **decode_state.kv_cache,
-            'input_pos': positions,
-        },
+        decode_cache_update_inputs,
         self.decode_cache_update_runner,
     )
 


john-rocky:
Added a main-vs-branch A/B to the description, and rebased onto main at 8379afb.

I ran Stage 2 twice against one Stage 1 bundle (Qwen3-0.6B, prefill 128 / cache 1024) with the same 3-prompt calibration set, changing only whether PYTHONPATH points at a pristine 8379afb tree or at this branch:

- before (main): Stage 2 stops after 2 s with `KeyError: 'valid_mask'` at `sampling_executor.py:525` in `prefill_chunk`, and writes no calibration output.
- after (this PR): calibrate finishes in 6 s (3/3 examples, 4 files written), Stage 3 quantize finishes in 7 s, and the SRQ bundle is written.

It reproduces on a plain macOS checkout with no NPU hardware involved — two `git archive` extracts and two `PYTHONPATH=` runs.

john-rocky:
One more data point, from today: the preset that ships in this repo hits it too.

I ran `npu_export/configs/gemma3_270m_qualcomm_sm8850.json` (`google/gemma-3-270m-it`) unmodified against main at `8379afb`. Stage 1 exports in 46 s. Stage 2 then stops on the first calibration prompt with the same `KeyError: 'valid_mask'`, at `sampling_executor.py:873` in `try_get_quantized_input`. With this branch applied to the same tree, and the same preset, Stage 2 finishes in 19 s and Stage 3 in 4 s.

The bundle from the patched run compiles through Stage 4 (QAIRT 2.47, SM8850) and runs on a Galaxy S26: four subgraphs dispatched to the NPU, 17.1k tok/s prefill at 128 tokens, 166 tok/s decode.

So the failure isn't specific to a config I put together — the gemma3-270m preset checked into this repo can't reach Stage 3 on main.

john-rocky:
Closing this: main now has the same fix. #1169 (d49f1ba, merged Aug 24) guards the same two cache-update call sites, and is the stronger change — it generalizes the batch dimension rather than assuming batch 1, and adds sampling_executor_test.py. I verified the #1177 repro passes at d49f1ba, so there is nothing left for this PR to do. Thanks for resolving it upstream.

Issue #1181, opened 2026-08-22
npu_export Stage 2 calibrates non-Gemma models with Gemma's chat markers, so the quantized model degrades on chat-templated prompts


john-rocky:
Thanks for keeping the NPU export pipeline moving. With #1177's `valid_mask` fix applied locally, Stages 1-3 run end to end on Qwen/Qwen3-0.6B from a plain checkout. One thing I ran into after that is worth reporting: the exported bundle generates fluent text from a bare prompt, but degrades sharply once the chat template is applied — which is what LiteRT-LM feeds at inference.

It traces to Stage 2. `quant_utils.get_example_prompt()` wraps plain-text calibration examples in `PROMPT_TEMPLATE_PREFIX` / `PROMPT_TEMPLATE_SUFFIX`, which are hardcoded to Gemma's `<start_of_turn>` markers. For a Qwen3 tokenizer those are not control tokens at all: `'<start_of_turn>user\n'` encodes to `[27, 2468, 3575, 37274, 29, 872, 198]`, ordinary text. Calibration therefore never observes `<|im_start|>` (151644) or `<|im_end|>` (151645), which are exactly what the runtime sends.

I should be precise about how far I can take that. What is measured is that the calibration string is the only variable: four routes that differ only in what string reaches the calibrator produce two outcomes, and the Gemma-marker one is the outlier. What I have not measured is the step from "different quantization bounds" to "degenerate text" — the bounds differ in 4398 of 4868 entries, but they do not widen or saturate (median width ratio between the two calibrations is 1.0000, p10 0.944, p90 1.093). So treat the mechanism as inference and the input dependence as the finding.

I've opened #1182 with the change and a regression test; take it in whatever form is easiest to import.

**How the default path reaches it**

[...]

PR #1182, opened 2026-08-22
Calibrate with the model's own chat template instead of Gemma's turn markers

john-rocky:
This is the fix proposed in #1181: Stage 2 calibration wraps plain-text examples in `PROMPT_TEMPLATE_PREFIX` / `PROMPT_TEMPLATE_SUFFIX`, which are Gemma's `<start_of_turn>` markers, so a non-Gemma model is calibrated on strings containing none of its own control tokens.

`get_example_prompt()` already calls `apply_chat_template()` for `{'messages': [...]}` examples. This routes the four plain-text branches — a bare string, `'text'`, `'prompt'`, `'inputs'` — through the same call via a new `_wrap_one_prompt()`, and keeps the constants as the fallback when no chat template is available. Callers that pass no tokenizer are unchanged.

**Gemma is unchanged at the token level.** The rendered string gains a leading `<bos>`, which `tokenizer.py` already de-duplicates, so token ids are identical on all 50 prompts in `sample_calibration_prompts.json` for both `gemma-3-270m-it` and `gemma-2-2b-it`.

**Verified as a bit-identical result against the branch that already works.** Three runs on a pristine `upstream/main` at `8379afb`, with only #1178 applied on top so Stage 2 can run at all. All three reuse one Stage 1 bundle (`Qwen/Qwen3-0.6B`, `prefill_lengths=[128]`, `cache_length=1024`, `dynamic_wi8_afp32`) and the same 10 prompts from `sample_calibration_prompts.json` at 32 decode steps.

| calibration data | quant_utils | quantized tflite sha256 | generation on the chat-templated prompt |
|---|---|---|---|
| `{"text": p}` | main | `a0b8ef0249d4c7f176b3671e597ba85b` | `<think>者
</think>or
 一个在不同国家和文化中...` |
| `{"messages": [...]}` | main | `86481cad55d7a874e64faf85e401e591` | `<think>
Okay, the user wants me to introduce myself briefly. Let me start by recalling my name and role. I` |
| `{"text": p}` | this branch | `86481cad55d7a874e64faf85e401e591` | `<think>
Okay, the user wants me to introduce myself briefly. Let me start by recalling my name and role. I` |

Rows 2 and 3 agree byte for byte. The change makes the shipped default data format reach the path that already produced a correct model; it does not introduce a new one.

Generation is greedy, 24 new tokens, from the 14 token ids the Jinja template produces for "Hello, briefly introduce yourself.". Stage 2 took 33 s and Stage 3 7 s in each of the three runs.

**Two guards, both triggered by real templates in my local set.** `apply_chat_template()` is wrapped in `try`/`except` and the result is checked for the prompt before it is used; either failure falls back to the constants with a warning rather than calibrating on a bad string.

- SmolVLM2's template expects list-shaped content and returns a user turn with the prompt dropped. Calibrating on 50 empty user turns would be worse than the current behaviour.
- The Qwen3.5 and Qwen3.8 templates raise `TemplateError: No user query found in messages.` when a calibration example starts with `<tool_response>` and ends with `</tool_response>`.

A non-string prompt still raises `TypeError` as it does on main, rather than silently producing an empty example.

**Tests.** `quant_utils_test.py` is new — `calib/` had no test file, and `run_tests.sh` collects it without registration. 12 tests; 6 fail on main and all pass here. The failing six are the ones asserting that a non-Gemma model's own template is used.

Not included, as a separate concern: `calib/tokenizer.py` doesn't pass `chat_template` on the `--transformers_model_path` branch, so a repo that stores its template in `chat_template.json` still lands on the fallback. `export_lib.py:308-323` already carries that workaround.

Measured at Stage 3, on the quantized tflite. I haven't re-run Stage 4 for the corrected bundle, so there's no device number here.

Happy to add the `ci:run` label myself if that's the convention here, or to reshape this into whatever form is easiest to import.

### DIFF
--- a/litert_torch/generative/export_hf/experimental/calib/quant_utils.py
+++ b/litert_torch/generative/export_hf/experimental/calib/quant_utils.py
@@ -46,9 +46,11 @@ class gfile:
   @staticmethod
   def IsDirectory(path): return os.path.isdir(path)
 
-# Following BARD250 format.
+# Following BARD250 format. These are Gemma's turn markers; they are only a
+# fallback for calibration runs that have no chat template to work from.
 PROMPT_TEMPLATE_PREFIX = '<start_of_turn>user\n'
 PROMPT_TEMPLATE_SUFFIX = '<end_of_turn>\n<start_of_turn>model\n'
+_warned_no_chat_template = False
 
 BASE_SAVE_DIR = ''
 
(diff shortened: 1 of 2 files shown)


john-rocky:
Pushed 81cf98e, which fixes a defect in what I sent first.

The original revision guarded the template call by checking that the rendered string still contained the prompt. That passes for the wrong reason when the prompt is a substring of the template's own boilerplate. With HuggingFaceTB/SmolVLM2-256M-Video-Instruct, a calibration example of `User:` rendered to a turn carrying no content, the check saw `User:` in the template's own `User: <end_of_utterance>` and accepted it — silently shortening that sample from 28 tokens to 9. `Assistant:`, `<|im_start|>` and `end_of_utterance` behaved the same way.

It now renders a sentinel through the template once and requires it to survive, so a template that drops plain-string content is caught regardless of what the prompt looks like. There is a test for that case; it fails on the previous revision.

The same push restores the original behaviour of the `{'messages'}` fallback when no chat template is available — routing it through the new helper changed that path, which was not part of this fix.

The equivalence result is unchanged: with the shipped `{"text": p}` data this branch still produces a quantized tflite byte-identical to the `{'messages'}` arm, sha256 86481cad55d7a874e64faf85e401e591.

aravindmurali22:
Hi @john-rocky, thank you for the contribution! Closing this PR as the primary fix was already imported and merged in #1193, and the remaining sentinel-probe follow-up is being tracked in #1197.

Issue #1194, opened 2026-08-25
export_hf sets start_token from tokenizer.bos_token even when add_bos_token is False — the bundle prepends a BOS the HF side never feeds (bos == eos checkpoints can degenerate into echoing the question)


john-rocky:
Thanks for the `export_hf` pipeline — it handled `ibm-granite/granite-4.1-3b` (3B dense) out of the box: the released pip stack exported, quantized, and ran on the first attempt. The one thing between that export and a clean gate turned out to be a single metadata field: `build_llm_metadata` writes `start_token` from `tokenizer.bos_token` even when the tokenizer declares `add_bos_token: false`, so the bundle prepends a BOS the same checkpoint never sees under HF. For granite, whose BOS is also its EOS, the first-turn output degenerates into echoing the question.

**Where it happens.** The write is unconditional on a truthy `bos_token` ([core/litert_lm_builder.py#L133-L139 at current main](https://github.com/google-ai-edge/litert-torch/blob/95ccef1b0f20d77495783a2695ca2b3bc64b4ac5/litert_torch/generative/export_hf/core/litert_lm_builder.py#L133-L139)):

```python
if hasattr(tokenizer, 'bos_token') and tokenizer.bos_token:
    ...
    llm_metadata.start_token.token_str = tokenizer.bos_token
```

`add_bos_token` is not consulted here or anywhere else under `export_hf`. A declared `bos_token` does not mean HF ever feeds it: granite-4.1 sets `add_bos_token: false` in `tokenizer_config.json` and has no post-processor in `tokenizer.json`, so HF-side tokenization never prepends anything. The runtime honours the metadata as declared — `MaybeGetBosString` / `ApplyPromptTemplates` in [runtime/core/session_utils.cc](https://github.com/google-ai-edge/LiteRT-LM/blob/38390524aba95b145ef55aa8934c7cd642dc8edf/runtime/core/session_utils.cc) prepend the declared start token on the first turn, in both the templated and non-templated branches — so the very first token the model reads is `<|end_of_text|>` (id 100257), an end-of-text marker. (L73–L74 nearby already strip a template-leading `bos_token`, so the runtime owning the BOS looks intentional; the uncovered case is a tokenizer that declares a BOS it never actually prepends.)

[...]

PR #1198, opened 2026-08-27
Write start_token only when the model input actually starts with the BOS

john-rocky:
Refs #1194 — thanks for the quick look at the report and for inviting this PR.

`build_llm_metadata` wrote `start_token` from `tokenizer.bos_token` whenever the attribute was truthy. For a checkpoint that declares a BOS but is never fed one (granite-4.1: `add_bos_token: false`, BOS == EOS), the runtime then prepends the EOS to the first turn and the model can degenerate into echoing the question — measurements in #1194. This PR writes `start_token` only when the model input actually starts with the BOS.

**The gate** (`_tokenizer_prepends_bos`) probes two things, in order:

1. Plain tokenization — does `tokenizer('x').input_ids` start with `bos_token_id`? This is the probe from #1194; it covers `add_bos_token` and post-processor-managed BOS.
2. If not, the shipped chat template — does its rendered form, tokenized with `add_special_tokens=False`, start with `bos_token_id`? Some templates carry the BOS themselves (OLMo-2's begins with `{{ bos_token }}`; SmolLM2-Instruct's ChatML opens with `<|im_start|>`, which is its declared BOS). `parse_chat_template` strips that prefix — and the runtime renders the jinja `bos_token` variable as empty — on the assumption that `start_token` restores it, so these models must keep the field. The probe receives the template actually being packaged, so `jinja_chat_template_override` is honored.

A tokenizer that cannot be probed (calling it raises) keeps the previous behavior of trusting the declared `bos_token`.

**Why probe instead of reading `add_bos_token`.** The attribute is not ground truth: gemma-2 and gemma-3 tokenizers expose `add_bos_token == False` yet do prepend `<bos>` through their post-processor. An attribute-based gate would have dropped their `start_token`; the probe keeps it.

**Checked against real tokenizers** (transformers 5.15.1), comparing the gate's outcome with whether HF's own input starts with the BOS:

| tokenizer | HF input starts with BOS? | start_token after this PR |
|---|---|---|
| ibm-granite/granite-4.1-3b | no — `add_bos_token: false`, template starts `<\|start_of_role\|>` | removed (the #1194 fix) |
| HuggingFaceTB/SmolLM2-135M | no — BOS == EOS, never fed | removed |
| allenai/OLMo-2-0425-1B-Instruct | yes — template begins `{{ bos_token }}` | kept |
| HuggingFaceTB/SmolLM2-135M-Instruct | yes — ChatML template opens with the BOS | kept |
| google/gemma-3-270m | yes — tokenization prepends `<bos>` | kept |
| Llama-3.2-1B-Instruct (unsloth mirror) | yes — post-processor prepends | kept |
| Qwen/Qwen3-0.6B | no `bos_token` declared | unchanged (absent) |

The override path is covered too: a base checkpoint with no template of its own, packaged with a `{{ bos_token }}`-style override, keeps its `start_token`.

**Tests.** `litert_lm_builder_test.py` covers both directions: the BOS == EOS non-prepending shape must not get a `start_token` (fails on current main), and the prepending / template-carried / override shapes must keep it. One note: running the test file by path (`python .../litert_lm_builder_test.py`) trips a pre-existing name collision between `core/litert_lm_builder.py` and the `litert-lm-builder` pip package; `./run_tests.sh` and `python -m` invocations are unaffected.

Happy to reshape any of this — the probe, the fallback, or the tests — into whatever form is easiest to review.

### DIFF
--- a/litert_torch/generative/export_hf/core/litert_lm_builder_test.py
+++ b/litert_torch/generative/export_hf/core/litert_lm_builder_test.py
@@ -18,7 +18,7 @@ from absl.testing import absltest
 from absl.testing import parameterized
 from litert_torch.generative.export_hf.core import export_lib
 from litert_torch.generative.export_hf.core import exportable_module
-from litert_torch.generative.export_hf.core import litert_lm_builder
+from litert_torch.generative.export_hf.core import litert_lm_builder as litertlm_builder
 
 _EMPTY_CHAT_TEMPLATES = ((None, None), (None, None), (None, None))
 
@@ -88,7 +88,7 @@ class _FakeModel:
 
 
 def _build_llm_metadata(tokenizer, chat_templates=_EMPTY_CHAT_TEMPLATES):
-  return litert_lm_builder.build_llm_metadata(
+  return litertlm_builder.build_llm_metadata(
       source_model_artifacts=export_lib.SourceModelArtifacts(
           model=_FakeModel(),
           model_config=None,
@@ -110,7 +110,7 @@ class TokenizerPrependsBosTest(parameterized.TestCase):
     tokenizer = _FakeTokenizer(
         bos_token="<s>", bos_token_id=1, prepends_bos=True
     )
-    self.assertTrue(litert_lm_builder._tokenizer_prepends_bos(tokenizer))
+    self.assertTrue(litertlm_builder._tokenizer_prepends_bos(tokenizer))
 
   def test_false_when_declared_bos_is_not_prepended(self):
     # add_bos_token: false checkpoints (e.g. granite-4.1) declare a BOS that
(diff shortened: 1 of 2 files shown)


sirakiin (dismissed):
Thanks for the PR!

litert_torch/generative/export_hf/core/litert_lm_builder_test.py
```diff
@@ -0,0 +1,261 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Tests for litert_lm_builder."""
+
+from absl.testing import absltest
+from absl.testing import parameterized
+from litert_torch.generative.export_hf.core import export_lib
+from litert_torch.generative.export_hf.core import exportable_module
+from litert_torch.generative.export_hf.core import litert_lm_builder
```
sirakiin: Nit: update to `import litert_lm_builder as litertlm_builder` to make it consistant with other files

john-rocky: Done in debde61 — the import is now aliased as `litertlm_builder` to match the other files. Thanks for the review!

sirakiin (dismissed):

aravindmurali22:
Closing this PR as it has already been merged into `main` via Copybara in commit 404270bd2a411c95dbca92955c42de8dfadfcc26. Thank you for contributing, @john-rocky!

